PR #5, opened 2024-07-23
Add typography for swiss transfer

LunarX:


sirambd (changes requested):
Add the default text style

sirambd (approved):

PR #8, opened 2024-07-26
feat: Adaptive two pane for Tablet and Mobile

sirambd:


app/src/main/java/com/infomaniak/swisstransfer/ui/components/TwoPaneScaffold.kt
```diff
@@ -0,0 +1,85 @@
...
+ * between list and detail panes.
+ *
+ * @param T The type parameter representing the data model used in the navigator.
+ * @param windowAdaptiveInfo An instance of [WindowAdaptiveInfo] that provides information about
+ * the current window's size class and adaptive state.
+ * @param listPane A composable function that describes the content of the list pane.
+ * It is provided with an instance of [ThreePaneScaffoldNavigator].
+ * @param detailPane A composable function that describes the content of the detail pane.
+ * It is provided with an instance of [ThreePaneScaffoldNavigator].
+ * @param modifier A [Modifier] for this composable. Defaults to [Modifier].
```
LunarX, line 45: Is all of these details really needed? Because most of these parameter descriptions are self explanatory, the comment is pretty much redundant and adds a lot of unnecessary noise in the file. 

The only parameter that could make sens to explain would be the generic type T. But even then, this parameter T could probably use a little bit more explanations as it only states what's written in the signature. Currently, it doesn't help understand what T's purpose is which should probably be the comment's main point.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/MainNavHost.kt
```diff
@@ -19,30 +19,44 @@
...
+import com.infomaniak.swisstransfer.ui.screen.main.sent.SentScreen
+import com.infomaniak.swisstransfer.ui.screen.main.settings.SettingsScreenWrapper
+import com.infomaniak.swisstransfer.ui.screen.main.transferdetails.TransferDetailsScreen
 
 @Composable
-fun MainNavHost(navController: NavHostController, startDestination: SentDestination) {
+fun MainNavHost(
+    navController: NavHostController,
+    startDestination: SentDestination,
+    windowAdaptiveInfo: WindowAdaptiveInfo,
```
LunarX, line 39: Do we always need to pass down the WindowAdaptiveInfo instance through all functions? Is there no way to get access it globally inside whichever function needs it?

LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/TwoPaneScaffold.kt
LunarX: By the way, what's the purpose of this T value?

LunarX (approved):

PR #9, opened 2024-07-26
core: Init new transfer UI screens

sirambd:
Depends on #8


app/src/main/java/com/infomaniak/swisstransfer/ui/BaseActivity.kt
```diff
@@ -0,0 +1,31 @@
...
+import androidx.activity.ComponentActivity
+import androidx.activity.enableEdgeToEdge
+
+open class BaseActivity : ComponentActivity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        enableEdgeToEdge()
+    }
+}
```
LunarX, line 31: Is it that useful to have a common BaseActivity? Is the only thing we're gonna share this `enableEdgeToEdge`? Because we will maybe need different colors for the bottom system bar in each activity in the future

LunarX (changes requested):

LunarX (approved):

PR #10, opened 2024-07-26
Add palette of colors used inside the app and start using some of them

LunarX:
This PR defines the main color palette of basic colors as defined inside the spec. 

It also starts defining some of the material 3 color roles. The choice is mainly based on our bottom navigation bar and a filled card colors in the specs when compared to the color role that they are supposed to use by default inside material 3.

A custom colors class is defined in order to access all the specific color correspondances that do not fit inside the material 3 color role system.

Finally, all these colors and applied to the navigation bar to match the spec.


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/MainScaffold.kt
```diff
@@ -65,6 +72,16 @@ private fun MainScaffold(
     currentDestination: MainNavigation,
     navController: NavHostController,
     content: @Composable () -> Unit,
+) {
+    Navigation(currentDestination, navController::navigateToSelectedItem, navType, content)
+}
```
sirambd, line 77: Why create a method to call another method ?

app/src/main/java/com/infomaniak/swisstransfer/ui/theme/Theme.kt
```diff
@@ -62,4 +82,7 @@ object SwissTransferTheme {
     val typography: CustomTypography
         @Composable
         get() = LocalCustomTypography.current
+    val colors: CustomColors
```
sirambd, line 85: Add delegate material colors

app/src/main/java/com/infomaniak/swisstransfer/ui/theme/ColorLight.kt
```diff
@@ -0,0 +1,72 @@
...
+
+val LightColors = CustomColors(
+    primaryTextColor = Color(orca),
+    secondaryTextColor = Color(elephant),
+    navigationItemBackground = light_background,
+    divider = Color(mouse),
+)
+
+@Immutable
+data class CustomColors(
```
sirambd, line 67: Move to Theme and add default color as Color.Unspecified

app/src/main/java/com/infomaniak/swisstransfer/ui/theme/Theme.kt
```diff
@@ -25,30 +25,50 @@ import androidx.compose.material3.darkColorScheme
...
+    onTertiary = light_onTertiary,
     background = light_background,
+    surface = light_surface,
+    onSurface = light_onSurface,
+    onSurfaceVariant = light_onSurfaceVariant,
+    surfaceContainerHighest = light_onSurfaceContainerHighest,
 )
 
 val LocalCustomTypography = staticCompositionLocalOf { Typography }
+val LocalCustomColors: ProvidableCompositionLocal<CustomColors?> = staticCompositionLocalOf { null }
```
sirambd, line 60: Use default instance

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/MainScaffold.kt
```diff
@@ -75,12 +92,28 @@ private fun MainScaffold(
...
-        content = content,
+        content = {
+            if (navType == NavigationSuiteType.None) {
+                content()
+            } else {
+                Column {
+                    Box(modifier = Modifier.weight(1f)) {
+                        content()
+                    }
+                    HorizontalDivider(color = SwissTransferTheme.colors.divider)
```
sirambd, line 113: Can we use the surface color here ?

app/src/main/java/com/infomaniak/swisstransfer/ui/theme/Theme.kt
```diff
@@ -62,4 +82,7 @@ object SwissTransferTheme {
     val typography: CustomTypography
         @Composable
         get() = LocalCustomTypography.current
+    val colors: CustomColors
+        @Composable
+        get() = LocalCustomColors.current ?: error("Cannot access uninitialized LocalCustomColors")
```
sirambd, line 87: Avoid runtime exceptions

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/MainScaffold.kt
```diff
@@ -65,6 +72,16 @@ private fun MainScaffold(
     currentDestination: MainNavigation,
     navController: NavHostController,
     content: @Composable () -> Unit,
+) {
+    Navigation(currentDestination, navController::navigateToSelectedItem, navType, content)
+}
+
+@Composable
+private fun Navigation(
```
sirambd, line 80: You can just directly update `MainScaffold` and add this callback instead of another method

sirambd (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/MainScaffold.kt
```diff
@@ -110,3 +143,16 @@ private fun NavHostController.navigateToSelectedItem(destination: MainNavigation
         restoreState = true
     }
 }
+
+@PreviewMobile
+@Composable
+private fun NavigationPreview() {
```
sirambd, line 149: Add preview tablet

app/src/main/java/com/infomaniak/swisstransfer/ui/theme/Theme.kt
LunarX: We can't use delegation on composable method but I added the property anyway

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/MainScaffold.kt
LunarX: The surface color wouldn't match the spec's colors. Why did you want to use surface specifically?

app/src/main/java/com/infomaniak/swisstransfer/ui/theme/Theme.kt
LunarX: Fixed by using the default unspecified value for the CustomColors class

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/MainScaffold.kt
LunarX: Fixed in https://github.com/Infomaniak/android-SwissTransfer/pull/10#discussion_r1694758215

sirambd (approved):

PR #23, opened 2024-08-16
Add a scaffold for all the screens that use a top app bar + sticky buttons at the bottom of the screen and some LargeButton component

LunarX:
This way we will always create the screens the same way.

This PR also adds the LargeButton component but later on this LargeButton should be defined with the more normal buttons which share a bunch of characteristics as well

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/ButtonType.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/ButtonType.kt
@@ -1,0 +1,80 @@
+/*
+ * Infomaniak SwissTransfer - Android
+ * Copyright (C) 2024 Infomaniak Network SA
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/components/ButtonType.kt
```diff
@@ -0,0 +1,80 @@
...
+import androidx.compose.ui.unit.dp
+import com.infomaniak.swisstransfer.ui.theme.Margin
+import com.infomaniak.swisstransfer.ui.theme.Shapes
+import com.infomaniak.swisstransfer.ui.theme.SwissTransferTheme
+
+@Composable
+fun LargeButton(
+    modifier: Modifier = Modifier,
+    @StringRes titleRes: Int,
+    style: ButtonType,
```
sirambd, line 41: Add primary as default type

sirambd (changes requested):

sirambd (approved):

PR #25, opened 2024-08-19
Add bottom sheet items component and use them in transfer upload source choice bottom sheet

LunarX:
Depends on #24

PR #26, opened 2024-08-19
Add placeholders settings for the main setting screen

LunarX:
Depends on #25

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/SettingsComponents.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/SettingsComponents.kt
@@ -37,18 +37,16 @@ import com.infomaniak.swisstransfer.ui.icons.app.Add
 import com.infomaniak.swisstransfer.ui.icons.app.ChevronRightSmall
 import com.infomaniak.swisstransfer.ui.icons.app.Folder
 import com.infomaniak.swisstransfer.ui.icons.app.OpenOutside
+import com.infomaniak.swisstransfer.ui.theme.Dimens
 import com.infomaniak.swisstransfer.ui.theme.Margin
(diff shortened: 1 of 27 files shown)


tevincent (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SettingsComponents.kt
sirambd, line 1: These components are settings-specific, so there's no need to share them globally, just create a components package in settings components and add them.
For example, the file: `.ui.screen.main.settings.components.SettingItem.kt` will contain `SettingItem` and `SettingItemContent`.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/SettingsScreen.kt
```diff
@@ -60,7 +76,35 @@ fun SettingsScreenWrapper(
 
 @Composable
 private fun SettingsScreen(onItemClick: (Any) -> Unit) {
-    Text("Settings screen")
+    Column(modifier = Modifier.verticalScroll(rememberScrollState())) {
+        Text(
```
sirambd, line 80: This is just for the poster example, but later it will be a toolbar, right?

sirambd (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/SettingsScreen.kt
LunarX: Yes, but I decided to not bother with adding the toolbar and the scaffold for now

tevincent (approved):

sirambd (approved):

PR #29, opened 2024-08-20
Add transfer types four big buttons

LunarX:
Depends on #27


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/TransferTypeButtons.kt
sirambd, line 1: Refactor as discussed

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/TransferTypeButtons.kt
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+@Composable
+private fun TransferTypeButton(
+    modifier: Modifier = Modifier,
+    label: String,
+    background: Color,
+    foreground: Color,
+    icon: ImageVector,
```
sirambd, line 72: Instead of passing all parameters, you can just pass the item

sirambd (changes requested):

PR #46, opened 2024-09-03
Format code

KevinBoulongne:
Depends on #44

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/Buttons.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/Buttons.kt
@@ -68,7 +68,7 @@ private fun CoreButton(
     enabled: Boolean,
     onClick: () -> Unit,
     imageVector: ImageVector?,
-    titleRes: Int
+    titleRes: Int,
 ) {
     Button(
         modifier = modifier.height(buttonSize.height),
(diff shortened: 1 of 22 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/components/Buttons.kt
```diff
@@ -69,7 +69,7 @@ private fun CoreButton(
     enabled: Boolean,
     onClick: () -> Unit,
     imageVector: ImageVector?,
-    titleRes: Int
+    titleRes: Int,
```
NicolasBourdin88, line 72: Maybe add @StringRes as the `SmallButton` function.

You should also maybe keep the same order for parameters on both function (`SmallButton() `and `LargeButton().`

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/sent/SentScreen.kt
```diff
@@ -95,18 +95,18 @@ fun EmptyScreen() {
...
             )
             NewTransferFab(
                 modifier = Modifier
                     .constrainAs(fab) {
-                        start.linkTo(parent.start)
-                        end.linkTo(parent.end)
-                        top.linkTo(parent.top, Margin.Large)
+                        start.linkTo(anchor = parent.start)
+                        end.linkTo(anchor = parent.end)
+                        top.linkTo(anchor = parent.top, margin = Margin.Large)
```
NicolasBourdin88, line 109: Is the `margin =` really necessary ?

NicolasBourdin88 (approved):

PR #63, opened 2024-09-18
Display UploadSourceChoiceBottomSheet before navigating to NewTransferScreen

KevinBoulongne:
Depends on #60

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/NewTransferFab.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/NewTransferFab.kt
@@ -20,24 +20,31 @@ package com.infomaniak.swisstransfer.ui.components
 import androidx.compose.material3.FloatingActionButtonDefaults
 import androidx.compose.material3.FloatingActionButtonElevation
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/components/NewTransferFab.kt
```diff
@@ -20,24 +20,31 @@ package com.infomaniak.swisstransfer.ui.components
...
+    var showUploadSourceChoiceBottomSheet by rememberSaveable { mutableStateOf(false) }
+
     SwissTransferFab(
         modifier = modifier,
         fabType = newTransferFabType.fabType,
         elevation = newTransferFabType.elevation(),
-        onClick = { context.launchActivity(NewTransferActivity::class) },
+        onClick = { showUploadSourceChoiceBottomSheet = true },
+    )
+    UploadSourceChoiceBottomSheet(
```
tevincent, line 45: Maybe we can move the `if` outside `UploadSourceChoiceBottomSheet` ? Because it's weird that `UploadSourceChoiceBottomSheet` is handling it's own visibility.

tevincent (changes requested):

tevincent (approved):

PR #66, opened 2024-09-19
Fix settings option style

LunarX:
This PR does the following:
* updates the font style of options to match the spec
* update the padding of option screen's gray subtitles
* simplify option screen definition by introducing a scaffold
* fix option's height when they have an icon or when they don't


tevincent (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/SettingsDownloadsLimitScreen.kt
```diff
@@ -17,26 +17,13 @@
  */
 package com.infomaniak.swisstransfer.ui.screen.main.settings
```
KevinBoulongne, line 18: I updated the Settings too in another PR, so I'm looking at this one first, to see if it will cause issues with my PR.
After that, I'll accept this one then rebase mine afterward.

KevinBoulongne (changes requested):

KevinBoulongne (approved):

sirambd (changes requested):

sirambd (approved):

PR #67, opened 2024-09-19
Fix top app bar

LunarX:
Fixes the text style of the SwissTransferTobAppBar and the disappearing back arrow when navigating back from the settings

Depends on #66

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/components/OptionScaffold.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/components/OptionScaffold.kt
@@ -44,12 +44,12 @@ fun OptionScaffold(
     navigateBack: (() -> Unit)? = null,
 ) {
     Scaffold(topBar = {
-        val canDisplayBackButton = if (LocalNavType.current == NavigationSuiteType.NavigationBar) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/components/OptionScaffold.kt
```diff
@@ -42,7 +44,11 @@ fun OptionScaffold(
     navigateBack: (() -> Unit)? = null,
 ) {
     Scaffold(topBar = {
-        val canDisplayBackButton = navigateBack?.let { TopAppBarButton.backButton(navigateBack) }
+        val canDisplayBackButton = if (LocalNavType.current == NavigationSuiteType.NavigationBar) {
```
KevinBoulongne, line 47: Why this TopAppBarButton is named like a boolean ?

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/components/OptionScaffold.kt
LunarX: Yeah, this makes no sens, I renamed it

KevinBoulongne (approved):

PR #78, opened 2024-09-25
core: Pooling all adaptive tablet logic

sirambd:
Depends on #57

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/utils/WindowAdaptativeInfoExt.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/utils/WindowAdaptativeInfoExt.kt
@@ -1,0 +1,46 @@
+/*
+ * Infomaniak SwissTransfer - Android
+ * Copyright (C) 2024 Infomaniak Network SA
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.infomaniak.swisstransfer.ui.utils
+
+import androidx.compose.material3.adaptive.WindowAdaptiveInfo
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/utils/WindowAdaptativeInfoExt.kt
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Determines if the current window is classified as a small window suitable for mobile devices.
+ *
+ * A window is considered small if its width size class is not [WindowWidthSizeClass.EXPANDED]
+ * or its height size class is [WindowHeightSizeClass.COMPACT]. This function can be used
+ * to adapt the UI layout for smaller screens, ensuring a better user experience on mobile devices.
+ *
+ * @return `true` if the window is small (mobile), `false` otherwise.
+ */
```
LunarX, line 45: The KDoc here doesn't bring any meaningful info, it adds more noise than it helps in my opinion. It's faster and more straight to the point to read the code directly.

The only interesting info could be the very first line of each KDoc that talks about tablets and phones, and maybe we could add examples of use cases for these conditions in the project (TwoPaneScaffold, top app bars, etc.)

LunarX (changes requested):

KevinBoulongne (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/sent/SentListScreen.kt
```diff
@@ -21,22 +21,23 @@ import androidx.compose.foundation.layout.padding
...
 import com.infomaniak.swisstransfer.ui.theme.SwissTransferTheme
 import com.infomaniak.swisstransfer.ui.utils.PreviewMobile
 import com.infomaniak.swisstransfer.ui.utils.PreviewTablet
+import com.infomaniak.swisstransfer.ui.utils.isWindowSmall
 
 @Composable
-fun SentListScreen(
-    navType: NavigationSuiteType,
-) {
+fun SentListScreen() {
```
KevinBoulongne, line 35: The app changed a lot in another currently open PR _(#57)_, so it will possibly easier to rebase this one upon the other one.
I'll try to do that.

In the meantime, I'm blocking this PR.

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/sent/SentListScreen.kt
KevinBoulongne: It's been rebased, all is good.

KevinBoulongne (approved):

LunarX (approved):

PR #82, opened 2024-10-08
Add Splashscreen & app icon

KevinBoulongne:


app/src/main/AndroidManifest.xml
```diff
@@ -34,7 +33,7 @@
         <activity
             android:name=".ui.MainActivity"
             android:exported="true"
-            android:theme="@style/Theme.AndroidSwissTransfer">
+            android:theme="@style/Theme.AppStarting">
```
LunarX, line 36: I'm not sure we want to call the theme of the main activity "AppStarting" no? Maybe it made sens when it was a separated activity on other apps or something but here I have doubts. What do you think?

app/build.gradle.kts
```diff
@@ -97,6 +97,7 @@ dependencies {
 
     // Others
     implementation(libs.kotlinx.serialization)
+    implementation(libs.androidx.core.splashscreen)
```
LunarX, line 100: In the end we need this library? Because last time we talked didn't we end up seeing that the library wasn't necessary and a single line with methods coming for the default android packages was enough?

app/src/main/res/drawable/ic_launcher_background.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?><!--
+<!--
```
LunarX, line 1: You removed mutliple xml headers, was it on purpose?

LunarX (changes requested):

app/src/main/AndroidManifest.xml
KevinBoulongne: This is the SplashScreen theme only, it's replaced by the `Theme.AndroidSwissTransfer` theme when the SplashScreen is done.

app/build.gradle.kts
KevinBoulongne: In kMail it's not used because we have a separated Activity to handle the SplashScreen.

In SwissTransfer, the MainActivity is handling the SplashScreen, so we need this library to be able to set the correct theme when the SplashScreen is done.

app/src/main/res/drawable/ic_launcher_background.xml
KevinBoulongne: The XML header is optional, and AndroidStudio never adds it when creating a new file, so I decided to just remove it.

app/src/main/AndroidManifest.xml
KevinBoulongne: _(With `postSplashScreenTheme` in `themes.xml`.)_

app/src/main/res/drawable/ic_launcher_background.xml
KevinBoulongne: This diff has been reverted.

LunarX (approved):

PR #84, opened 2024-10-09
Create a file tile component

LunarX:
For now the FileTile component relies on a temporary FileUiItem while we wait for the KMP model

PR #85, opened 2024-10-09
Create a FileTileList component

LunarX:
This is reused in a multiple screens so here is an easy to use component to reuse it everywhere. For now, multiselection is handled in the received screen as a demonstration

Depends on #84


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/sent/SentViewModel.kt
```diff
@@ -28,4 +30,5 @@ import javax.inject.Inject
 @HiltViewModel
 class SentViewModel @Inject constructor() : ViewModel() {
     val transfers = flow<List<Any>> { emit(listOf(1)) }.stateIn(viewModelScope, SharingStarted.Eagerly, null)
+    val selectedTransferIds: SnapshotStateMap<String, Boolean> = mutableStateMapOf<String, Boolean>()
```
sirambd, line 33: Don't expose mutable state

sirambd (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/sent/SentViewModel.kt
LunarX: Yeah there doesn't seem to be any non mutable one

sirambd (approved):

PR #87, opened 2024-10-14
Adds and use the component to display picked files during a new transfer

LunarX:
For now it displays picked files without copying them to the app's local storage. This will come in a second PR

Depends on #85

PR #88, opened 2024-10-14
Use a PreviewParameter to factorize FileUiItem instantiation inside previews

LunarX:
I was always using the same list of FileUiItems so instead of copy pasting it everywhere, I used the PreviewParameter that serves this very purpose

Depends on #87

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/FileItem.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/FileItem.kt
@@ -27,7 +27,9 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.tooling.preview.PreviewParameter
 import androidx.compose.ui.unit.dp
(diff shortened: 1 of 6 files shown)


KevinBoulongne (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SmallFileTile.kt
```diff
@@ -65,32 +67,22 @@ enum class SmallFileTileSize(val size: Dp, val shape: Shape) {
 @Preview(name = "Light")
 @Preview(name = "Dark", uiMode = Configuration.UI_MODE_NIGHT_YES or Configuration.UI_MODE_TYPE_NORMAL)
 @Composable
-private fun SmallFileTilePreview() {
+private fun SmallFileTilePreview(@PreviewParameter(FileUiListPreviewParameter::class) files: List<FileUiItem>) {
     SwissTransferTheme {
         Surface(color = SwissTransferTheme.materialColors.surfaceContainerHighest) {
             Column(Modifier.padding(16.dp)) {
+                val file = files[0]
```
sirambd, line 74:
```suggestion
                val file = files.first()
```

sirambd (changes requested):

sirambd (approved):

PR #90, opened 2024-10-15
Add Matomo & Sentry opt-out related screens

KevinBoulongne:
Depends on #86


LunarX (changes requested):
I propose that the whole feature be in its own Android module so it can be reused in all future apps easily

tevincent (changes requested):
Maybe we can remove DATA_SOURCE_CODE ?

tevincent (approved):

PR #103, opened 2024-10-17
refactor: Typography is now an object

KevinBoulongne:


LunarX (commented):
I agree with the idea, but yeh let's wait and hear @sirambd 's opnion

tevincent (approved):

sirambd (approved):

PR #104, opened 2024-10-17
refactor: Rename CustomShapes variables

KevinBoulongne:


LunarX (changes requested):
I agree that we need all five values to either start with an uppercase or a lowercase. I don't know which one is better though, I'd wait on @sirambd if he has an opinion on it.

Still, can't we do anything in order to name it Shapes instead of CustomShapes?

tevincent (approved):

sirambd (approved):

PR #115, opened 2024-10-22
feat: Add transfer advanced settings

FabianDevel:


KevinBoulongne (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/TransferAdvancedSettings.kt
```diff
@@ -0,0 +1,79 @@
...
+
+@Composable
+fun TransferAdvancedSettings(states: () -> List<TransferAdvancedOptionsEnum>, onClick: () -> Unit) {
+    SwissTransferCard(modifier = Modifier.padding(Margin.Medium)) {
+        TransferAdvancedSettingType.entries.forEach { settingType ->
+            val title by remember { derivedStateOf { states()[settingType.ordinal] } }
+            TransferAdvancedSetting(settingType, { title }, onClick)
+        }
+    }
+}
```
LunarX, line 54: I'd still extract the modifier outside the TransferAdvancedSettings component because the component should not have to know from where it's being called. At least these paddings.

LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SharpRippleButton.kt
```diff
@@ -33,7 +33,8 @@ fun SharpRippleButton(
     modifier: Modifier = Modifier,
     isSelected: () -> Boolean = { false },
     onClick: () -> Unit,
-    content: @Composable RowScope.() -> Unit,
+    contentPadding: PaddingValues = PaddingValues(0.dp),
+    content: @Composable (RowScope.() -> Unit),
```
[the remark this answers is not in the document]
FabianDevel: Good catch, I don't even know how the ended up there in the first place 🤔

app/src/main/java/com/infomaniak/swisstransfer/ui/images/icons/Password.kt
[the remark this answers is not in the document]
FabianDevel: It came directly from the figma icon, it's now updated

app/src/main/java/com/infomaniak/swisstransfer/ui/images/icons/LockedTextFieldSmall.kt
LunarX, line 1: Try to import the "original" icon's version not the "flattened" one

LunarX, line 1: "Small" in the name does not qualify icons in 16x16 but icons with thicker lines. Also the name "small" is not correct I think, we should rather stick with the name "thick" which is also already used. But in the case of this icon, there no need for any small nor thick

LunarX (changes requested):

LunarX (approved):

PR #117, opened 2024-10-23
feat: Add FilesDetailScreen

tevincent:
Depends on https://github.com/Infomaniak/android-SwissTransfer/pull/326

https://github.com/user-attachments/assets/2af6036f-207e-4c61-9ca1-7ff32bc7eb31

Depends on https://github.com/Infomaniak/multiplatform-SwissTransfer/pull/81

Screen to display details of files provided by the parameter `filesList`.

Can be configured like this:
- `folderUuid`: folder UUID in which we want to navigate
- `navigateToFolder`: this callback is called when the user click on a file
- `withFileSize`: if we want to display files size
- `withSpaceLeft`: if we want to display the space left, useful when displaying files of a transfer
- `withFileDelete`: if deletion of files is enable
- `navigateBack`: this callback is called when the user want to navigate back


app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTopAppBar.kt
```diff
@@ -31,6 +31,40 @@ import com.infomaniak.swisstransfer.ui.images.icons.Cross
...
+            titleContentColor = SwissTransferTheme.colors.toolbarTextColor,
+            actionIconContentColor = SwissTransferTheme.colors.toolbarIconColor,
+            navigationIconContentColor = SwissTransferTheme.colors.toolbarIconColor,
+        ),
+        title = { Text(text = title, style = SwissTransferTheme.typography.h2) },
+        navigationIcon = { navigationMenu?.let { MenuButton(navigationMenu) } },
+        actions = { actionMenus.forEach { actionMenu -> MenuButton(actionMenu) } },
+    )
+}
+
```
LunarX, line 67: Don't duplicate the code like this, define the inner component as taking in a String and if given a stringRes in the outer component, call the inner component by reading the string ressource. This way both cases call the same bit of code. Always.

Always try to make it so the similar code like this (where both cases should always be the same) is defined depending on one another so nobody can ever introduce any form of difference between the two even if they tried

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/MainNavHost.kt
```diff
@@ -54,6 +55,19 @@ fun MainNavHost(
             val transferDetails: TransferDetailsDestination = it.toRoute()
             TransferDetailsScreen(transferId = transferDetails.transferId)
         }
+        composable<FilesDetailsDestination> {
+            val filesDetailsDestination: FilesDetailsDestination = it.toRoute()
+            FilesDetailsScreen(
+                navigateToDetails = { fileId ->
+                    //TODO check here if specified filedId is a folder
```
LunarX, line 62: What's the temporality of this TODO?

app/src/main/java/com/infomaniak/swisstransfer/ui/utils/HumanReadableSizeUtils.kt
```diff
@@ -24,10 +24,20 @@ import androidx.compose.runtime.Composable
...
 object HumanReadableSizeUtils {
-    fun getHumanReadableSize(context: Context, sizeInBytes: Long): String {
-        return Formatter.formatShortFileSize(context, sizeInBytes)
+    private const val TOTAL_FILE_SIZE: Long = 50_000_000_000L
+
+    private fun getHumanReadableSize(context: Context, sizeInBytes: Long) = Formatter.formatFileSize(context, sizeInBytes)
+
+    private fun getFilesSizeInBytes(files: List<FileUi>) = files.sumOf { it.fileSizeInBytes }
+
+    fun Context.getFilesSize(files: List<FileUi>): String = getHumanReadableSize(this, getFilesSizeInBytes(files))
```
LunarX, line 36: Maybe we should call it something along the lines of `getSpaceUsed` to match the `getSpaceLeft`? If "FilesSize" is more meaningful than "Space", we can also replace space with file size, what do you prefer?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/FilesDetailsViewModel.kt
```diff
@@ -0,0 +1,54 @@
...
+            ),
+            FileUi(
+                fileName = "The 5 step guide to turning it off and on again.docx",
+                uid = "The 5 step guide to turning it off and on again.docx",
+                fileSizeInBytes = 89723143,
+                mimeType = null,
+                uri = "",
+            ),
+        )
+    }
```
LunarX, line 53: Maybe already make this into a flow even if we hardcode its values

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/FilesSize.kt
```diff
@@ -0,0 +1,83 @@
...
+import com.infomaniak.swisstransfer.ui.utils.HumanReadableSizeUtils.formatSpaceLeft
+import com.infomaniak.swisstransfer.ui.utils.HumanReadableSizeUtils.getFilesSize
+import com.infomaniak.swisstransfer.ui.utils.HumanReadableSizeUtils.getSpaceLeft
+
+@Composable
+fun FilesSize(files: List<FileUi>, withSpaceLeft: Boolean) {
+    Row(
+        modifier = Modifier
+            .padding(vertical = Margin.Medium),
+    ) {
```
LunarX, line 46: I'd leave matters such as overall display choices (here paddings) to the calling parent and have the component be "call site agnostic" if you get what I mean

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/FilesSize.kt
```diff
@@ -0,0 +1,83 @@
...
+        val filesCount = files.count()
+        val filesSize = LocalContext.current.getFilesSize(files)
+        val filesDetail = "${pluralStringResource(R.plurals.filesCount, filesCount, filesCount)}  •  $filesSize"
+        Text(
+            filesDetail,
+            modifier = Modifier
```
LunarX, line 52: I don't know if you prefer it on two lines or if you didn't know, but you don't have to put it on multiple lines if it's as short as this line

app/src/main/java/com/infomaniak/swisstransfer/ui/utils/HumanReadableSizeUtils.kt
```diff
@@ -24,10 +24,20 @@ import androidx.compose.runtime.Composable
...
 import androidx.compose.ui.res.pluralStringResource
 import com.infomaniak.swisstransfer.R
+import com.infomaniak.swisstransfer.ui.components.FileUi
 
 object HumanReadableSizeUtils {
-    fun getHumanReadableSize(context: Context, sizeInBytes: Long): String {
-        return Formatter.formatShortFileSize(context, sizeInBytes)
+    private const val TOTAL_FILE_SIZE: Long = 50_000_000_000L
+
+    private fun getHumanReadableSize(context: Context, sizeInBytes: Long) = Formatter.formatFileSize(context, sizeInBytes)
```
LunarX, line 32: Why did the format change from formatShortFileSize to formatFileSize?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/FilesSize.kt
```diff
@@ -0,0 +1,83 @@
...
+import com.infomaniak.swisstransfer.ui.components.FileUi
+import com.infomaniak.swisstransfer.ui.previewparameter.FileUiListPreviewParameter
+import com.infomaniak.swisstransfer.ui.theme.Margin
+import com.infomaniak.swisstransfer.ui.theme.SwissTransferTheme
+import com.infomaniak.swisstransfer.ui.utils.HumanReadableSizeUtils.formatSpaceLeft
+import com.infomaniak.swisstransfer.ui.utils.HumanReadableSizeUtils.getFilesSize
+import com.infomaniak.swisstransfer.ui.utils.HumanReadableSizeUtils.getSpaceLeft
+
+@Composable
+fun FilesSize(files: List<FileUi>, withSpaceLeft: Boolean) {
```
LunarX, line 42: Is it fine to not use a lambda for files here?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/NewTransferNavHost.kt
```diff
@@ -52,5 +54,17 @@ fun NewTransferNavHost(navController: NavHostController, closeActivity: () -> Un
...
             UploadSuccessScreen(transferType = TransferType.MAIL)
         }
+        composable<FilesDetailsDestination> {
+            val filesDetailsDestination: FilesDetailsDestination = it.toRoute()
+            FilesDetailsScreen(
+                navigateToDetails = { fileId ->
+                    navController.navigate(FilesDetailsDestination(fileId))
+                },
+                fileId = filesDetailsDestination.fileId,
+                withSpaceLeft = false, onCloseClicked = {},
```
LunarX, line 64: You've placed two arguments on the same line

LunarX, line 64: Can't we already code `onCloseClicked`?

LunarX, line 64: Shouldn't withSpaceLeft be true here?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/FilesDetailsScreen.kt
```diff
@@ -0,0 +1,108 @@
...
+            files = filesToDisplay,
+            isRemoveButtonVisible = onFileRemoved != null,
+            isCheckboxVisible = false,
+            isUidChecked = { false },
+            setUidCheckStatus = { _, _ -> },
+            onRemoveUid = { onFileRemoved?.invoke(it) },
+            onClick = { navigateToDetails(it) }
+        )
+    }
+}
```
LunarX, line 91: I don't know to what extent this part is optimized for recompositions. If you haven't already, maybe just make sure it's optimized with @sirambd or by checking the recompositions in the layout inspector. And if you do ask him, I'd be interested to listen to the explanation as well btw

LunarX (changes requested):
Is this screen being used in another PR already? If it is, then it's fine like this but if it's not, we should use the screen already for the imported files during a new transfer, it could help see some details and issue we might not have thought about

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/NewTransferNavHost.kt
```diff
@@ -52,5 +54,17 @@ fun NewTransferNavHost(navController: NavHostController, closeActivity: () -> Un
...
+                onFileRemoved = {},
+                navigateBack = { navController.popBackStack() }
```
KevinBoulongne, line 66:
```suggestion
                navigateToDetails = { fileId -> navController.navigate(FilesDetailsDestination(fileId)) },
                fileId = filesDetailsDestination.fileId,
                withSpaceLeft = false,
                onCloseClicked = {},
                onFileRemoved = {},
                navigateBack = { navController.popBackStack() },
```

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/FilesSize.kt
```diff
@@ -0,0 +1,83 @@
...
+            )
+        }
+    }
+}
+
+@Preview(name = "Light")
+@Preview(
+    name = "Dark",
+    uiMode = Configuration.UI_MODE_NIGHT_YES or Configuration.UI_MODE_TYPE_NORMAL,
+)
```
KevinBoulongne, line 75:
```suggestion
@Preview(name = "Dark", uiMode = Configuration.UI_MODE_NIGHT_YES or Configuration.UI_MODE_TYPE_NORMAL)
```

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/FilesSize.kt
```diff
@@ -0,0 +1,77 @@
...
+import com.infomaniak.swisstransfer.ui.utils.HumanReadableSizeUtils.getSpaceLeft
+import com.infomaniak.swisstransfer.ui.utils.HumanReadableSizeUtils.getSpaceUsed
+
+@Composable
+fun FilesSize(modifier: Modifier, files: () -> List<FileUi>, withSpaceLeft: Boolean) {
+    Row(modifier = modifier) {
+        val files = files()
+        val filesCount = files().count()
+        val filesSize = LocalContext.current.getSpaceUsed(files)
+        val filesDetail = "${pluralStringResource(R.plurals.filesCount, filesCount, filesCount)}  •  $filesSize"
```
KevinBoulongne, line 47: This string was already created in `ImportedFilesCard`, but in a different way _(by using 3 Texts instead of concatenating them)_.
I think we should factorize the usage of `•` in the project, because I'm also adding it in 2 more places in another PR.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/FilesDetailsViewModel.kt
```diff
@@ -0,0 +1,60 @@
...
+import kotlinx.coroutines.flow.stateIn
+import javax.inject.Inject
+
+@HiltViewModel
+class FilesDetailsViewModel @Inject constructor() : ViewModel() {
+
+    fun getFilesFromUUID(uuid: String): Flow<List<FileUi>?> {
+        // TODO Call the right manager to fetch files from Realm
+        val fileUiList = listOf(
+            FileUi(
```
KevinBoulongne, line 36: Since this PR is using the `FileUi` class, I think it should be rebased upon #136 before merging.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/FilesDetailsScreen.kt
```diff
@@ -0,0 +1,109 @@
...
+private fun FilesDetailsScreenPreview() {
+    SwissTransferTheme {
+        Surface {
+            FilesDetailsScreen(
+                fileId = "",
+                navigateToDetails = { },
+                withSpaceLeft = true,
+                onFileRemoved = { },
+                onCloseClicked = { },
+                navigateBack = { },
```
KevinBoulongne, line 105:
```suggestion
                navigateToDetails = {},
                withSpaceLeft = true,
                onFileRemoved = {},
                onCloseClicked = {},
                navigateBack = {},
```

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/FilesSize.kt
KevinBoulongne: Seen with @LunarX.
I'm doing a Component for this, you'll be able to use it after that.

KevinBoulongne (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/MainNavHost.kt
```diff
@@ -48,6 +50,19 @@ fun MainNavHost(
...
+                    navController.navigate(FilesDetailsDestination(fileId))
+                },
+                fileId = filesDetailsDestination.fileId,
+                navigateBack = { navController.popBackStack() },
+                withSpaceLeft = false,
+                onCloseClicked = {},
```
LunarX, line 63: Do we not have a "onCloseClicked" yet?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/FilesSize.kt
```diff
@@ -0,0 +1,76 @@
...
+fun FilesSize(modifier: Modifier, files: () -> List<FileUi>, withSpaceLeft: Boolean) {
+    Row(modifier = modifier) {
+        val context = LocalContext.current
+        TextDotText(
+            firstText = {
+                val filesCount = files().count()
+                pluralStringResource(R.plurals.filesCount, filesCount, filesCount)
+            },
+            secondText = { HumanReadableSizeUtils.getHumanReadableSize(context, files().sumOf { it.fileSize }) },
+            modifier = Modifier.padding(start = Margin.Medium),
```
LunarX, line 52: The standard way to add spacing, like here, in jetpack compose is to use a `Spacer()` instead of specifying a padding like you did. There's plenty in the code you can look at to see how it's done elsewhere

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/FilesDetailsScreen.kt
```diff
@@ -0,0 +1,109 @@
...
+fun FilesDetailsScreen(
+    filesDetailsViewModel: FilesDetailsViewModel = hiltViewModel<FilesDetailsViewModel>(),
+    fileId: String,
+    navigateToDetails: (String) -> Unit,
+    withSpaceLeft: Boolean,
+    onFileRemoved: ((uuid: String) -> Unit)? = null,
+    onCloseClicked: (() -> Unit),
+    navigateBack: (() -> Unit),
+) {
+    val files by filesDetailsViewModel.getFilesFromUUID(fileId).collectAsStateWithLifecycle(emptyList())
```
LunarX, line 47: We'll need to be able to get a list of files for the root of all files as well, not just for a given UUID. 

But this can be implemented at the same time as the logic behind getFilesFromUUID I guess

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/FilesDetailsScreen.kt
```diff
@@ -0,0 +1,109 @@
...
+    navigateBack: (() -> Unit),
+) {
+    Column {
+        SwissTransferTopAppBar(
+            title = title(),
+            navigationMenu = TopAppBarButton.backButton { navigateBack() },
+            TopAppBarButton.closeButton { onCloseClicked() },
+        )
+
+        FilesSize(Modifier.padding(vertical = Margin.Medium), files, withSpaceLeft)
```
LunarX, line 80: Same here about paddings and `Spacer()`s. Maybe is that one more debatable?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/FilesDetailsScreen.kt
```diff
@@ -0,0 +1,109 @@
...
+        withSpaceLeft = withSpaceLeft,
+        onFileRemoved = onFileRemoved,
+        onCloseClicked = onCloseClicked,
+        navigateBack = navigateBack,
+    )
+}
+
+@Composable
+private fun FilesDetailsScreen(
+    title: () -> String,
```
LunarX, line 65: The title doesn't look like it needs to be a state, can it change value?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/FilesDetailsScreen.kt
```diff
@@ -0,0 +1,109 @@
...
+    files: () -> List<FileUi>,
```
LunarX, line 66: Same question about files actually. Will the list of file items change?

LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/FilesSize.kt
LunarX: Yeah we can put the whole "modifier" line in a single line. It's usually useful when there's only one modifier to apply

app/src/main/java/com/infomaniak/swisstransfer/ui/utils/HumanReadableSizeUtils.kt
LunarX: I asked the UI/UX, he wanted the same logic as the web version where there's only 1 decimal number and if the decimal is 0 as in "50.0 GB" the decimal would be gone. "50.00 GB" felt kind of cluttered compared to "50 GB" for him.

I don't know what's the closest we can get to what he wants?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/NewTransferNavHost.kt
LunarX: Yes, I think so too, definitely

tevincent: As we discussed with @sirambd and you, we'll do nothing with the cross as it adds a lot of complexity to the app to remove all the stack of `FileDetailsScreen` depending of where we came from.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/FilesDetailsScreen.kt
tevincent: I talked about that with @sirambd. We'll keep the modifier but we'll use it inside `FileSize` since the `paddingVertical` is not changing.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/MainScreen.kt
```diff
@@ -40,11 +43,42 @@ fun MainScreen() {
         derivedStateOf { navBackStackEntry?.toMainDestination() ?: MainNavigation.startDestination }
     }
 
+    var lastStartDestination: MainNavigation by remember { mutableStateOf(MainNavigation.startDestination) }
```
sirambd, line 46: Use directly a callback from the screen, and each destination can set its own popbackstack

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/MainScreen.kt
```diff
@@ -40,11 +43,42 @@ fun MainScreen() {
...
+
     MainScaffold(
         navController = navController,
         currentDestination = currentDestination,
-        largeWindowTopAppBar = { BrandTopAppBar() },
-        content = { MainNavHost(navController, currentDestination) },
+        windowTopAppBar = { isWindowLarge ->
```
sirambd, line 51: use `LocalWindowAdaptiveInfo.current` instead

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/MainNavHost.kt
```diff
@@ -32,23 +32,59 @@ import com.infomaniak.swisstransfer.ui.navigation.MainNavigation.*
...
+                },
+                folderUuid = filesDetailsDestination.folderUuid,
+                navigateBack = { navController.popBackStack() },
+                close = {
+                    closeFilesDetails()
+                },
+                withFilesSize = false,
+                withSpaceLeft = false,
+                withFileDelete = false,
+                shouldDisplayTopAppBar = isWindowSmall,
```
sirambd, line 86: You can use the `LocalWindowAdaptiveInfo.current`

sirambd (dismissed):

tevincent:
What I've done:

- Hide the space left compose element instead of passing an empty string
- `FilesDetailsScreen` has been split into two components depending of where they are used. `FilesDetailsScreen` for the one used in the `MainNavHost` and `NewTransferFilesDetailsScreen` for the one used in the new transfer screen
- Destination has also been renamed accordingly to their screens
- Loading of files has been modified by you

What's left:
- Implementing the callback in ReceivedScreen and SentScreen instead of having a variable to retain where we came from.

sirambd:
> This has been reviewed with @sirambd. Waiting for his approval.
> 
> What I've done:
> 
> * Hide the space left compose element instead of passing an empty string
> * `FilesDetailsScreen` has been split into two components depending of where they are used. `FilesDetailsScreen` for the one used in the `MainNavHost` and `NewTransferFilesDetailsScreen` for the one used in the new transfer screen
> * Destination has also been renamed accordingly to their screens
> * Loading of files has been modified by you
> 
> What's left:
> 
> * Implementing the callback in ReceivedScreen and SentScreen instead of having a variable to retain where we came from.

It seems to me that you were saying that with `popBackStack` solution it doesn't work using `previousBackStackEntry` but i don't know which case you were saying it doesn't work in.

tevincent:
> > This has been reviewed with @sirambd. Waiting for his approval.
> > What I've done:
> > 
> > * Hide the space left compose element instead of passing an empty string
> > * `FilesDetailsScreen` has been split into two components depending of where they are used. `FilesDetailsScreen` for the one used in the `MainNavHost` and `NewTransferFilesDetailsScreen` for the one used in the new transfer screen
> > * Destination has also been renamed accordingly to their screens
> > * Loading of files has been modified by you
> > 
> > What's left:
> > 
> > * Implementing the callback in ReceivedScreen and SentScreen instead of having a variable to retain where we came from.
> 
> It seems to me that you were saying that with `popBackStack` solution it doesn't work using `previousBackStackEntry` but i don't know which case you were saying it doesn't work in.

If it's working for you, I might have tried the wrong way. We can go with your solution.

sirambd:
> > > This has been reviewed with @sirambd. Waiting for his approval.
> > > What I've done:
> > > 
> > > * Hide the space left compose element instead of passing an empty string
> > > * `FilesDetailsScreen` has been split into two components depending of where they are used. `FilesDetailsScreen` for the one used in the `MainNavHost` and `NewTransferFilesDetailsScreen` for the one used in the new transfer screen
> > > * Destination has also been renamed accordingly to their screens
> > > * Loading of files has been modified by you
> > > 
> > > What's left:
> > > 
> > > * Implementing the callback in ReceivedScreen and SentScreen instead of having a variable to retain where we came from.
> > 
> > 
> > It seems to me that you were saying that with `popBackStack` solution it doesn't work using `previousBackStackEntry` but i don't know which case you were saying it doesn't work in.
> 
> If it's working for you, I might have tried the wrong way. We can go with your solution.

I'll wait for you to come back, so we can test it together and confirm whether it really works or not.
In the meantime, I'll draft it.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/components/FilesDetailsScreen.kt
```diff
@@ -0,0 +1,105 @@
...
+    files?.let {
+        SwissTransferScaffold(
+            topBar = {
+                SwissTransferTopAppBar(
+                    navigationIcon = { TopAppBarButtons.Back(onClick = navigateBack) },
+                    actions = { TopAppBarButtons.Close(close) },
+                )
+            },
+        ) {
+            Column {
```
LunarX, line 53: Do you need that column?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/components/FilesDetailsScreen.kt
```diff
@@ -0,0 +1,105 @@
...
+import com.infomaniak.swisstransfer.ui.screen.main.transferdetails.TransferDetailsViewModel
+import com.infomaniak.swisstransfer.ui.screen.newtransfer.FilesSize
+import com.infomaniak.swisstransfer.ui.theme.Margin
+import com.infomaniak.swisstransfer.ui.theme.SwissTransferTheme
+import com.infomaniak.swisstransfer.ui.utils.PreviewAllWindows
+
+@Composable
+fun FilesDetailsScreen(
+    paddingValues: PaddingValues = PaddingValues(0.dp),
+    transferDetailsViewModel: TransferDetailsViewModel = hiltViewModel<TransferDetailsViewModel>(),
```
LunarX, line 28: Ok I just realized that this component depends on a view model. Do we want a reusable component to depend on a view model? 

Now I'm a bit lost as to what is used where actually. Maybe let's just see this together live. Maybe I misunderstood some things in this file and this might invalid some of my previous comments

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transfers/TransfersScreenWrapper.kt
```diff
@@ -132,20 +144,71 @@ private fun DetailPane(
     navigator: ThreePaneScaffoldNavigator<DestinationContent>,
     hasTransfer: () -> Boolean,
 ) {
-
     val destinationContent = navigator.safeCurrentContent()
 
     if (destinationContent == null) {
         NoSelectionEmptyState(hasTransfer())
+    } else if (destinationContent.folderUuid != null) {
```
LunarX, line 151: Can this be turned into a `== null` check? Also, would it be easier to read as a `when` expression?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transfers/TransfersScreenWrapper.kt
```diff
@@ -132,20 +144,71 @@ private fun DetailPane(
...
+@OptIn(ExperimentalMaterial3AdaptiveApi::class)
+@Composable
+private fun FilesDetailsScreen(
+    windowAdaptiveInfo: WindowAdaptiveInfo,
+    navigator: ThreePaneScaffoldNavigator<DestinationContent>,
+    folderUuid: String,
+    transferDirection: TransferDirection,
+    transferUuid: String,
+    navigateToFolder: (folderUuid: String) -> Unit,
+) {
```
LunarX, line 192: Is this the counterpart of the FilesDetailsScreen that uses a view model declared in FilesDetailsScreen.kt?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/NewTransferFilesDetailsScreen.kt
```diff
@@ -0,0 +1,105 @@
...
+    onFileRemoved: ((uuid: String) -> Unit)? = null,
+    navigateBack: (() -> Unit),
+) {
+    Scaffold(
+        topBar = {
+            SwissTransferTopAppBar(
+                navigationIcon = { TopAppBarButtons.Back(onClick = navigateBack) },
+            )
+        }
+    ) { paddingValues ->
```
LunarX, line 80: Are we not missing a close button on the top right of the top app bar? (With a cross)

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transfers/TransfersScreenWrapper.kt
```diff
@@ -132,20 +144,71 @@ private fun DetailPane(
...
+        folderUuid = folderUuid,
+        navigateBack = { navigator.popBackStack() },
+        close = {
+            if (windowAdaptiveInfo.isWindowSmall()) navigator.navigateBack()
+            navigator.navigateToDetails(
+                windowAdaptiveInfo,
+                transferDirection,
+                transferUuid
+            )
+        },
```
LunarX, line 206: I'm not sure to get what this does exactly but, does it not leave the whole FilesDetailsScreen flow altogether no matter how deep in folders you are when you click on the close button?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transfers/TransfersScreenWrapper.kt
```diff
@@ -169,6 +232,7 @@ private fun NoSelectionEmptyState(hasTransfers: Boolean) {
 private data class DestinationContent(
     val direction: TransferDirection,
     val transferUuid: String,
+    val folderUuid: String? = null,
```
LunarX, line 235: This might be more of a detail because as a matter of fact this still works correctly, but wouldn't a sealed class be more appropriate here to represent the two types of navigations? If I understood correctly we might either navigate to the list of all files where `folderUuid` has no value or the a specific `folderUuid` and in this case it's set?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/FilesSize.kt
LunarX, line 1: This component is only needed for the NewTransferFilesDetailsScreen right? If it's true, then it should be defined in a `component` package located inside a package specifically created for this `NewTransferFilesDetailsScreen()` like other screens have

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/FilesSize.kt
```diff
@@ -0,0 +1,99 @@
...
+import com.infomaniak.swisstransfer.ui.theme.Margin
+import com.infomaniak.swisstransfer.ui.theme.SwissTransferTheme
+import com.infomaniak.swisstransfer.ui.utils.HumanReadableSizeUtils
+import com.infomaniak.swisstransfer.ui.utils.HumanReadableSizeUtils.formatSpaceLeft
+import com.infomaniak.swisstransfer.ui.utils.HumanReadableSizeUtils.getSpaceLeft
+
+@Composable
+fun FilesSize(files: List<FileUi>, withFilesSize: Boolean, withSpaceLeft: Boolean) {
+    Row(modifier = Modifier.padding(vertical = Margin.Medium)) {
+        Spacer(Modifier.size(Margin.Medium))
```
LunarX, line 43: Defining this horizontal padding here with a spacer is none of this component's business. This should be defined through a modifier where this component is called instead.

It's used next to a `FileItemList()` and this component defines its own horizontal padding which NEEDS to be the same as in here. But defining the horizontal padding of `FilesSize()` not only makes the component less reusable but also hides the idea that the padding needs to stay the same by splitting this Margin.Medium value across multiple files

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/NewTransferNavHost.kt
```diff
@@ -43,12 +46,16 @@ fun NewTransferNavHost(
     NavHost(navController, NewTransferNavigation.startDestination) {
         composable<ImportFilesDestination> {
             ImportFilesScreen(
+                importFilesViewModel = hiltViewModel<ImportFilesViewModel>(it),
```
LunarX, line 49: Does it still make sense to keep the default value inside ImportFilesScreen() definition if we're always specifying it here instead now?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/NewTransferNavHost.kt
```diff
@@ -112,5 +119,15 @@ fun NewTransferNavHost(
                 },
             )
         }
+        composable<NewTransferFilesDetailsDestination> {
+            val backStackEntry = remember(it) { navController.getBackStackEntry(ImportFilesDestination) }
+            NewTransferFilesDetailsScreen(
+                importFilesViewModel = hiltViewModel<ImportFilesViewModel>(backStackEntry),
```
LunarX, line 125: Same question, does it make sense to keep the default `importFilesViewModel` value inside the method declaration?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/NewTransferNavHost.kt
```diff
@@ -112,5 +119,15 @@ fun NewTransferNavHost(
...
+                withFilesSize = true,
+                withSpaceLeft = true,
+                withFileDelete = true,
+                navigateBack = { navController.popBackStack() },
+            )
```
LunarX, line 130: Are we not missing the top right close button?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -244,7 +247,13 @@ private fun ImportFilesScreen(
...
                 val modifier = Modifier.padding(horizontal = HORIZONTAL_PADDING)
                 SendByOptions(modifier, selectedTransferType)
-                FilesToImport(modifier, files, navigateToFileDetails = { /*TODO*/ }, addFiles, shouldStartByPromptingUserForFiles)
+                FilesToImport(
+                    modifier = modifier,
+                    files = files,
+                    navigateToFilesDetails = navigateToFilesDetails,
+                    addFiles = addFiles,
+                    shouldStartByPromptingUserForFiles = shouldStartByPromptingUserForFiles,
+                )
```
LunarX, line 256: This more of a detail but because every name is exactly the same, I would've not specified the argument names, but this is pretty subjective

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -282,7 +291,12 @@ private fun FilesToImport(
     LaunchedEffect(Unit) { if (shouldShowInitialFilePick) pickFiles() }
 
     ImportFilesTitle(modifier, R.string.myFilesTitle)
-    ImportedFilesCard(modifier, files, ::pickFiles, navigateToFileDetails)
+    ImportedFilesCard(
+        modifier,
+        files,
+        ::pickFiles,
+        navigateToFilesDetails = { navigateToFilesDetails() }
```
LunarX, line 298: Could this have been passed down directly?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/ImportedFilesCard.kt
```diff
@@ -58,20 +58,19 @@ fun ImportedFilesCard(
...
     files: () -> List<FileUi>,
     pickFiles: () -> Unit,
-    navigateToFileDetails: () -> Unit,
+    navigateToFilesDetails: () -> Unit,
 ) {
 
     val context = LocalContext.current
-
-    val humanReadableSize by remember {
+    val humanReadableSize by remember(files()) {
```
LunarX, line 65: Was this remember key really needed? You're passing down a state all the way down here so the derivedState should handle it be itself

LunarX (approved):

PR #118, opened 2024-10-23
Copy & share transfer link

KevinBoulongne:


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/upload/UploadSuccessQrScreen.kt
```diff
@@ -41,8 +45,17 @@ import com.infomaniak.swisstransfer.ui.theme.Margin
...
 @Composable
-fun UploadSuccessQrScreen(transferType: TransferType) {
+fun UploadSuccessQrScreen(transferType: TransferType, transferLink: String = "https://chk.me/83azQOl") {
+
+    val context = LocalContext.current
+    val sendIntent = Intent(Intent.ACTION_SEND).apply {
+        putExtra(Intent.EXTRA_TEXT, transferLink)
+        type = "text/plain"
+    }
+    val shareIntent = Intent.createChooser(sendIntent, null)
```
LunarX, line 57: Put this logic in its own util file by exposing a `shareString` method or `shareTextPlain` or similar and its "copy to clipboard" alternative as well

LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/utils/Extensions.kt
```diff
@@ -0,0 +1,38 @@
...
+import android.content.Context
+import android.content.Intent
+import androidx.compose.ui.platform.ClipboardManager
+import androidx.compose.ui.text.AnnotatedString
+
+object Extensions {
+
+    fun ClipboardManager.copyText(text: String) {
+        setText(AnnotatedString(text))
+    }
```
LunarX, line 29: You can do something like this to avoid having to get the ClipboardManager inside the compose code. Also this will prevent developers from forgetting to send a feedback to the user after a successful copy

```suggestion
    fun Context.copyText(text: String, showSnackbar: (String) -> Unit) {
        val clipboardManager = getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
        clipboardManager.setPrimaryClip(ClipData.newPlainText(text, text))

        showSnackbar(getString(R.string.textCopiedTODO))
    }
```

LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/utils/Extensions.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    fun Context.shareText(text: String) {
+        val intent = Intent(Intent.ACTION_SEND).apply {
+            putExtra(Intent.EXTRA_TEXT, text)
+            type = "text/plain"
+        }
+        startActivity(Intent.createChooser(intent, null))
+    }
+}
```
LunarX, line 38: We avoid generic "Extensions" fonction, do like the other extensions methods

LunarX (changes requested):

LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/upload/UploadSuccessQrScreen.kt
```diff
@@ -62,12 +76,15 @@ fun UploadSuccessQrScreen(transferType: TransferType) {
...
             )
         },
-        content = { Content(transferType) },
+        content = { Content(context, snackbarHostState, transferType, transferLink) },
     )
 }
 
 @Composable
-private fun Content(transferType: TransferType) {
+private fun Content(context: Context, snackbarHostState: SnackbarHostState, transferType: TransferType, transferLink: String) {
```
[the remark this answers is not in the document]
KevinBoulongne: Google displays the snackbar in the onClick lambda, like us.

app/src/main/java/com/infomaniak/swisstransfer/ui/utils/ContextExt.kt
```diff
@@ -58,3 +58,30 @@ fun Context.openAppNotificationSettings() {
         }
     }.also(::startActivity)
 }
+
+fun Context.copyText(text: String, showSnackbar: (String) -> Unit) {
+    val clipboardManager = getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
+    clipboardManager.setPrimaryClip(ClipData.newPlainText(text, text))
+
+    showSnackbar(getString(R.string.snackbarLinkCopiedToClipboard))
```
FabianDevel, line 66: We only need to provide feedback for android version < 13, because the sytstem takes care of it starting from Android 13

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/utils/ContextExt.kt
LunarX: I don't know to what extent this applies to all manufacturer but the document does indeed seem to say that it's a standard so yeah I agree maybe we should skip it for API 13+: https://developer.android.com/develop/ui/views/touch-and-input/copy-paste#Feedback

LunarX (approved):

FabianDevel (approved):

PR #120, opened 2024-10-24
feat: Add `SwissTransferTextField`

FabianDevel:


app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTextField.kt
```diff
@@ -0,0 +1,177 @@
...
+import kotlinx.coroutines.flow.asStateFlow
+
+@Composable
+fun SwissTransferTextField(
+    modifier: Modifier = Modifier,
+    isRequired: Boolean = true,
+    label: String? = null,
+    initialValue: String = "",
+    minLineNumber: Int = 1,
+    maxLineNumber: Int = if (minLineNumber > 1) Int.MAX_VALUE else 1,
```
LunarX, line 57: Deriving the number of max lines from the number of min lines instinctively feels out of place and like it should not be dependent

app/src/main/res/values/strings.xml
```diff
@@ -82,11 +84,16 @@
...
+    <string name="textFieldOptional">(Optional)</string>
+    <string name="transferMessagePlaceholder">A message to pass on?</string>
+    <string name="transferRecipientAddressPlaceholder">Send the files to</string>
     <string name="transferSendButton">Send</string>
+    <string name="transferSenderAddressPlaceholder">Your email address</string>
     <plurals name="transferSpaceLeft">
         <item quantity="one">%s left</item>
         <item quantity="other">%s left</item>
     </plurals>
+    <string name="transferTitlePlaceholder">Title your transfer</string>
```
LunarX, line 96: I don't know what this was supposed to mean but the wording feels strange, does it not?

app/src/main/res/values/strings.xml
```diff
@@ -82,11 +84,16 @@
         <item quantity="one">%d day</item>
         <item quantity="other">%d days</item>
     </plurals>
+    <string name="textFieldOptional">(Optional)</string>
+    <string name="transferMessagePlaceholder">A message to pass on?</string>
```
LunarX, line 88: Is "pass on" the correct verb? Because I feel like it has a strange connotation for this context. Maybe something along the lines of "transmit" or similar other words would fit better?

app/src/main/res/values-es/strings.xml
```diff
@@ -77,11 +79,16 @@
...
+    <string name="textFieldOptional">(Opcional)</string>
+    <string name="transferMessagePlaceholder">¿Algún mensaje que quiera transmitir?</string>
+    <string name="transferRecipientAddressPlaceholder">Envíe los archivos a</string>
     <string name="transferSendButton">Enviar</string>
+    <string name="transferSenderAddressPlaceholder">Su dirección de correo electrónico</string>
     <plurals name="transferSpaceLeft">
         <item quantity="one">Queda %s</item>
         <item quantity="other">Quedan %s</item>
     </plurals>
+    <string name="transferTitlePlaceholder">Dé un título a su transferencia</string>
```
LunarX, line 91: The verbs are conjugated with the polite form instead of the second person singular form

app/src/main/res/values-de/strings.xml
```diff
@@ -77,11 +79,16 @@
         <item quantity="one">%d Tag</item>
         <item quantity="other">%d Tage</item>
     </plurals>
+    <string name="textFieldOptional">(Fakultativ)</string>
+    <string name="transferMessagePlaceholder">Eine Botschaft, die Sie weitergeben möchten?</string>
+    <string name="transferRecipientAddressPlaceholder">Senden Sie die Dateien an</string>
     <string name="transferSendButton">Senden</string>
+    <string name="transferSenderAddressPlaceholder">Ihre E-Mail Adresse</string>
```
LunarX, line 86: The strings are conjugated to the polite form and to the 2nd person plural form instead of the 2nd person singular form

app/src/main/java/com/infomaniak/swisstransfer/ui/images/icons/EyeSmall.kt
LunarX, line 1: Can you import an "original" form of this icon or does it not exist?

app/src/main/java/com/infomaniak/swisstransfer/ui/images/icons/EyeCrossedOutSmall.kt
LunarX, line 1: Can you import an "original" form of this icon or does it not exist?

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTextField.kt
```diff
@@ -0,0 +1,177 @@
...
+    keyboardType: KeyboardType = KeyboardType.Text,
+    errorMessageState: StateFlow<String?> = MutableStateFlow(null).asStateFlow(),
```
LunarX, line 59: This feels like passing an entier flow is not the correct way of doing things. We usually pass in callbacks to either read or update the state that's managed outside. Maybe the sole reason is lowering the coupling and keeping a high cohesion, I don't know.

If you want to exact explanation ask @sirambd and tell me if you go ask him because I'm interested as well

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTextField.kt
```diff
@@ -0,0 +1,177 @@
...
+    supportingText: String? = null,
+    onValueChange: ((String) -> Unit)? = null,
```
LunarX, line 61: Is onValueChange really needed?

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTextField.kt
```diff
@@ -0,0 +1,177 @@
...
+import com.infomaniak.swisstransfer.ui.images.icons.EyeCrossedOutSmall
+import com.infomaniak.swisstransfer.ui.images.icons.EyeSmall
+import com.infomaniak.swisstransfer.ui.theme.Margin
+import com.infomaniak.swisstransfer.ui.theme.SwissTransferTheme
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+
+@Composable
+fun SwissTransferTextField(
```
LunarX, line 51: Can you try to keep the all powerful SwissTransferTextField private and only expose two specific versions that limit the options for each use case? Also the specific logic should be extract as much as possible outside the all powerful SwissTransferTextField and defined in their own specific version

LunarX (changes requested):

app/src/main/res/values-it/strings.xml
LunarX, line 1: Italian strings have been updated to fix some wordings I think

LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTextField.kt
FabianDevel: It's actually quite logical when thought about.
Otherwise we need to add another parameter like the `singleLine` one in the original textField
or force to provides parameter, which is not a good thing.

app/src/main/res/values/strings.xml
FabianDevel: Well it wasn't from me and I understood it instinctly, so I guess it doesn't feel that strange
It's the string for the transfer title field's placeholder

app/src/main/res/values-de/strings.xml
FabianDevel: We don't use polite form in the app ? So uncivilised

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTextField.kt
FabianDevel: Yes thanks I actually thought it was stange

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTextField.kt
FabianDevel: Yes it's a mandatory parameter, we cannot keep it inside because the text will often be stored in a viewModel to check error

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTextField.kt
FabianDevel: I don't quite like all separating a thing as simple as a TextField, seems like over engineering to me.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -103,11 +112,13 @@ private fun ImportFilesScreen(
...
-        topButton = { modifier ->
-            SendButton({ isImporting }, { importProgress }, modifier)
-        },
+        topButton = { modifier -> SendButton({ isImporting }, { importProgress }, modifier) },
         content = {
-            Column(Modifier.padding(horizontal = Margin.Medium, vertical = Margin.Large)) {
+            Column(
+                modifier = Modifier
+                    .padding(horizontal = Margin.Medium, vertical = Margin.Large)
+                    .verticalScroll(rememberScrollState())
```
KevinBoulongne, line 120:
```suggestion
                    .verticalScroll(rememberScrollState()),
```

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -116,8 +127,40 @@ private fun ImportFilesScreen(
...
+                TransferAdvancedSettings(
+                    states = {
+                        listOf(
+                            ValidityPeriodOption.THIRTY,
+                            DownloadLimitOption.ONE,
+                            PasswordTransferOption.NONE,
+                            EmailLanguageOption.GERMAN,
+                        )
+                    },
+                    onClick = { },
```
KevinBoulongne, line 162:
```suggestion
                    onClick = {},
```

KevinBoulongne (changes requested):

app/src/main/res/values/strings.xml
FabianDevel: We search in dictionnaries, turned out pass on is in fact the correct verb

sirambd (approved):

app/src/main/res/values/strings.xml
FabianDevel: Turns out this wording shouldn't even exists because the Figma was wrong, the API doesn't allow this field.
So I removed this

LunarX: Wouldn't something like "Give a title to your transfer" feel more natural?

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTextField.kt
LunarX: If the goal is only to force a single line when none of the parameters are defined, why not simply default minLineNumber to 1 and default maxLineNumber to 1?

Right now, by default, our text component doesn't grow with the text content but suddenly, when I only specify a minLineNumber, it changes its behavior to start growing with the text. That's what feels strange to me

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -117,8 +128,21 @@ private fun ImportFilesScreen(
                     showUploadSourceChoiceBottomSheet = { showUploadSourceChoiceBottomSheet = true },
                     removeFileByUid = removeFileByUid,
                 )
-                ImportFilesTitle(titleRes = R.string.transferTypeTitle)
-                ImportFilesTitle(titleRes = R.string.advancedSettingsTitle)
+                ImportTextFields(initialShouldShowEmailAddressesFields)
+                ImportFilesTitle(Modifier.padding(vertical = Margin.Medium), titleRes = R.string.transferTypeTitle)
+                TransferTypeButtons(initialSelectedTransferType = TransferType.LINK, onClick = {})
```
LunarX, line 133: Your buttons have some kind of padding that cuts them before reaching the edge of the screen. The component should go up to the edge of the screen and take in some form of contentPadding to solve this issue the same way Material components do

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -130,6 +154,28 @@ private fun ImportFilesScreen(
     )
 }
 
+@Composable
+private fun ImportTextFields(initialShouldShowEmailAddressesFields: Boolean) {
```
LunarX, line 158: I don't think this boolean has anything to do with an "initial" value

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -104,11 +113,13 @@ private fun ImportFilesScreen(
...
-            SendButton({ isImporting }, { importProgress }, modifier)
-        },
+        topButton = { modifier -> SendButton({ isImporting }, { importProgress }, modifier) },
         content = {
-            Column(Modifier.padding(horizontal = Margin.Medium, vertical = Margin.Large)) {
+            Column(
+                modifier = Modifier
+                    .padding(horizontal = Margin.Medium, vertical = Margin.Large)
+                    .verticalScroll(rememberScrollState()),
+            ) {
```
LunarX, line 122: What do you think about splitting the `content` of the column inside 3 or 4 methods: MyFiles(), TextFields(), SendBy() and AdvancedSettings() or something similar?

LunarX (changes requested):

app/src/main/res/values/strings.xml
FabianDevel: We don't really care anymore, as I say this string shouldn't have existed in the first place. This feature don't exists

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTextField.kt
FabianDevel: because maxLines prevails on minLines
So if I do as you say, then when I want to use it I specify only minLines = 3
The textField will stay a single line
And if we make a multi line TextField, that's because we want a potentially long message so this is dumb to not let it grows.

app/src/main/java/com/infomaniak/swisstransfer/ui/images/icons/EyeCrossed.kt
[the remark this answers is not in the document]
FabianDevel: I originally used CrossedOut because that's litterally what the eye is, but the Ui use only Crossed, and that's also a understandable way to call it.
That eye is litterally crossed out, that's like the official name of this icon if you search it on the web.  Slash is more confusing I find as it doesn't really represent the icon

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
FabianDevel: Yeah it will be rework in the next PR with the states, for now it's temporary code for integration

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
FabianDevel: This should also be in a PR truly dedicated to the ImportScreen

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
FabianDevel: That's a good idea but it should be done in another PR really dedicated to the importScreen.
This one is a component PR, I integrated it in the screen quickly to see the results and put it in better perspective, but it wasn't meant to replace the real PR for the whole screen

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTextField.kt
LunarX: I agree that minLines = 3 giving a 1 line textfield is an issue but this could be fixed with the following.
```suggestion
    minLineNumber: Int = 1,
    maxLineNumber: Int = if (minLineNumber > 1) minLineNumber else 1,
```

The question at stake here is the "is going to grow automatically" property of a text field changing in, from my point of view, an unintuitive manner. To be sure to understand your point, you defend the idea of that property changing automatically to be better than having to consciously specify that you want to field to grow?

With all this talk I'm also wondering, would sticking to the native maxLine and minLine behavior be more simple and intuitive to use?

app/src/main/res/values/strings.xml
LunarX: Oh my bad, I didn't get what you meant, no problem then!

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
LunarX: I didn't understand that this PR doesn't include the whole screen, then I understand

app/src/main/java/com/infomaniak/swisstransfer/ui/images/icons/EyeCrossed.kt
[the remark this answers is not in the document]
LunarX: I see that "crossed" seems to indeed be the most commonly used term. 

I guess iOS used the term "slash" instead of "cross" to precisely list and describe the elements in the icon to follow the guideline. This helps avoid naming collision in case there's an eye icon with a cross on it.

I don't know if it's more important to stick to a convention if it makes the icon's name harder to find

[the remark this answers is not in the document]
LunarX: For now the name is fine imo and doesn't bloc the PR from being merged. We can discuss in team later on to choose the appropriate naming

LunarX (approved):

PR #125, opened 2024-10-29
feat: Add UI for received transfers screen

KevinBoulongne:
Depends on #118

The quality gate failure about duplicated code can be ignored, it's only about testing code.

PR #126, opened 2024-10-29
Expired transfer BottomSheet

KevinBoulongne:
Depends on #125

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/TextDotText.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/TextDotText.kt
@@ -35,6 +35,7 @@ fun TextDotText(
     modifier: Modifier = Modifier,
     color: Color = SwissTransferTheme.colors.secondaryTextColor,
     style: TextStyle = SwissTransferTheme.typography.bodySmallRegular,
+    optionalSecondTextColor: Color? = null,
 ) {
     Row(modifier) {
         CustomText(firstText, color, style)
@@ -45,7 +46,7 @@ fun TextDotText(
             style = style,
         )
         Spacer(Modifier.width(Margin.Mini))
-        CustomText(secondText, color, style)
+        CustomText(secondText, optionalSecondTextColor ?: color, style)
     }
 }
 
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferExpiredBottomSheet.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferExpiredBottomSheet.kt
@@ -61,7 +61,6 @@ fun TransferExpiredBottomSheet(
     }
 
     SwissTransferBottomSheet(
-        modifier = Modifier,
         onDismissRequest = closeBottomSheet,
         bottomButton = {
             LargeButton(
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferExpiredBottomSheet.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferExpiredBottomSheet.kt
@@ -24,6 +24,8 @@ import androidx.compose.material3.Surface
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.stringResource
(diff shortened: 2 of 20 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/components/TextDotText.kt
```diff
@@ -35,6 +35,7 @@ fun TextDotText(
     modifier: Modifier = Modifier,
     color: Color = SwissTransferTheme.colors.secondaryTextColor,
     style: TextStyle = SwissTransferTheme.typography.bodySmallRegular,
+    optionalSecondTextColor: Color? = null,
```
tevincent, line 38: Do we have to specify that this is optional since it's nullable ?

app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferExpiredBottomSheet.kt
```diff
@@ -0,0 +1,100 @@
...
+@Composable
+fun TransferExpiredBottomSheet(
+    isVisible: () -> Boolean,
+    expirationDate: () -> Date?,
+    downloadsLimit: () -> Int?,
+    onDeleteTransferClicked: () -> Unit,
+    closeBottomSheet: () -> Unit,
+) {
+
+    if (!isVisible()) return
```
tevincent, line 51: Should we have this condition outside this component ?

app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferExpiredBottomSheet.kt
```diff
@@ -0,0 +1,100 @@
...
+        stringResource(
+            R.string.transferExpiredDateReachedDescription,
+            date.format(FORMAT_DATE_SIMPLE),
+        )
+    } else {
+        stringResource(R.string.transferExpiredLimitReachedDescription, downloadsLimit()!!)
+    }
+
+    SwissTransferBottomSheet(
+        modifier = Modifier,
```
tevincent, line 64: Why do we pass a Modifier without any modifier here ?

app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItem.kt
```diff
@@ -45,6 +46,27 @@ import java.util.UUID
 @OptIn(ExperimentalLayoutApi::class)
 @Composable
 fun TransferItem(transfer: TransferUi, onClick: () -> Unit) {
+
+    val createdDate = Date(transfer.createdDateTimestamp).format(FORMAT_DATE_TITLE)
+    val expirationDate = Date(transfer.expirationDateTimestamp)
+    val remainingDays = transfer.expiresInDays
+    val remainingDownloads = transfer.downloadLeft
+    val uploadedSize = Formatter.formatShortFileSize(LocalContext.current, transfer.sizeUploaded)
```
tevincent, line 54: We should use the `HumanReadableSizeUtils` no ?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/received/ReceivedScreen.kt
```diff
@@ -157,7 +163,37 @@ private fun ReceivedScreen(areTransfersEmpty: () -> Boolean) {
...
+                        }
+                        else -> {
+                            Log.d("TODO", "Display transfer details screen")
+                            // TODO
+                        }
+                    }
+                }
+            )
+
+            TransferExpiredBottomSheet(
```
tevincent, line 184: Maybe, having a `if (isExpired)` instead of `isVisible` for the condition might be clearer. That way, we know the `BottomSheet` is only visible if the transfer is expired.

tevincent (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferExpiredBottomSheet.kt
KevinBoulongne: I was doing stuff with this Modifier before, but forgot to remove it when I stopped using it.

Nice catch !

tevincent (approved):

PR #135, opened 2024-10-30
fix: Update tablet preview with new api and remove deprecated config

sirambd:


app/src/main/java/com/infomaniak/swisstransfer/ui/utils/MultiplePreview.kt
```diff
@@ -57,14 +57,13 @@ annotation class PreviewSmallWindow
 )
 @Preview(
     name = "(3) Large window landscape light",
-    group = "LargeWindow",
-    device = "spec:id=reference_tablet,shape=Normal,width=1280,height=800,unit=dp,dpi=240",
+    group = "LargeWindow", device = "spec:width=1280dp,height=800dp,dpi=240",
```
LunarX, line 60: You placed both `group` and `device` arguments on the same line?

LunarX (changes requested):

LunarX (approved):

PR #141, opened 2024-11-01
refactor: Make Previews great again!

KevinBoulongne:
Depends on #138

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/components/TransferInfo.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/components/TransferInfo.kt
@@ -17,10 +17,13 @@
  */
 package com.infomaniak.swisstransfer.ui.screen.main.transferdetails.components
 
+import android.content.res.Configuration.UI_MODE_NIGHT_YES
+import android.content.res.Configuration.UI_MODE_TYPE_NORMAL
 import android.text.format.Formatter
 import androidx.compose.foundation.layout.*
 import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.Icon
+import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.derivedStateOf
@@ -32,6 +35,8 @@ import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.pluralStringResource
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.tooling.preview.PreviewParameter
 import com.infomaniak.multiplatform_swisstransfer.common.interfaces.ui.TransferUi
 import com.infomaniak.swisstransfer.R
 import com.infomaniak.swisstransfer.ui.components.TextDotText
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/previewparameter/ListPreviewData.kt
LunarX, line 1: You should keep the old file separation where each provider is defined in its own file and where each provider's "main" detail is also in its own file. If you need to reuse the list of FileUi inside the provider for the list of TransferUi, I'd expose the variable in the FileUi file and access it from the TransferUi file instead.

Same for the TransferUi, if you need to access the list outside the preview for temporary code, still define it in the provider's code and access it in the code from there.

Maybe add a TODO comment that explains to remove the exposed variable once all of the temporary code has been removed

LunarX, line 1: Also, I thought I had written a comment in the FileUi parameter provider but apparently I forgot. I needed the first two items to stay the same as previously because I needed index 0 to be a non-preview file (i.e. pdf, txt, etc.) and I needed index 1 to be a preview file (i.e. png, jpg, etc.). Could you put back the correct types and add a comment on both items specifying their type please?

LunarX (changes requested):
First comments, the rest will come later

app/src/main/java/com/infomaniak/swisstransfer/ui/previewparameter/ListPreviewData.kt
KevinBoulongne: The FileUi & TransferUi lists are used outside of the Providers scope, so they shouldn't be together.

app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItemList.kt
```diff
@@ -67,97 +66,10 @@ fun TransferItemList(
...
-            }.time.time,
-            sizeUploaded = 237_866_728L,
-            downloadLimit = 420_069,
-            downloadLeft = 402_690,
-            message = "3ème transfert. RAS.",
-            files = files,
-        ),
-    )
-
+private fun Preview(@PreviewParameter(TransferUiListPreviewParameter::class) transfers: List<TransferUi>) {
```
tevincent, line 72:
```suggestion
private fun TransferItemListPreview(@PreviewParameter(TransferUiListPreviewParameter::class) transfers: List<TransferUi>) {
```

tevincent (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItemList.kt
KevinBoulongne: Setting a customized name for each preview is kinda tedious.
With @FabianDevel, we would prefer to simply name them all `Preview()`.
@LunarX is kinda ok about that.

LunarX (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItemList.kt
LunarX: I'd say, if it's the standard way of coding in jetpack compose or the good practice to name them, then maybe we should keep the names. 

But if it's not that recommended then I don't care much either way as I use a live template to automatically generate the preview so that even if we need a name, the name is automatically filled in most of the time

tevincent (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItemList.kt
KevinBoulongne: After reading the documentation, it looks like there isn't any convention yet.

PR #143, opened 2024-11-01
feat: Add `PreviewLightAndDark` for Components

KevinBoulongne:
Depends on #141


tevincent (approved):

FabianDevel (changes requested):
We didn't do this initially because it puts the dark theme preview first
We should poll the teams to know if we want this or not

PR #142, opened 2024-11-01
feat: Add transfer button type state

FabianDevel:

PR #144, opened 2024-11-01
feat: Add advanced option buttons states

FabianDevel:
Depends on #142 

The password option is useless in this PR, but as it's not mandatory for a first working alpha, It will be developped it in a later PR

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/SettingsViewModel.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/SettingsViewModel.kt
@@ -62,13 +62,13 @@ class SettingsViewModel @Inject constructor(
     data class AppSettingsData(
         val validityPeriod: ValidityPeriod,
         val downloadLimit: DownloadLimit,
-        val isPasswordActivated: Boolean,
+        val passwordOption: PasswordTransferOption,
         val emailLanguage: EmailLanguage,
     ) {
         fun toAdvancedOptionsList() = listOf<TransferAdvancedOptionsEnum>(
             validityPeriod.toAdvancedOption(),
             downloadLimit.toAdvancedOption(),
-            if (isPasswordActivated) PasswordTransferOption.ACTIVATED else PasswordTransferOption.NONE,
+            passwordOption,
             emailLanguage.toAdvancedOption(),
         )
     }
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/SettingsViewModel.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/SettingsViewModel.kt
@@ -28,8 +28,8 @@ import com.infomaniak.swisstransfer.di.IoDispatcher
 import com.infomaniak.swisstransfer.ui.screen.main.settings.DownloadLimitOption.Companion.toAdvancedOption
 import com.infomaniak.swisstransfer.ui.screen.main.settings.EmailLanguageOption.Companion.toAdvancedOption
 import com.infomaniak.swisstransfer.ui.screen.main.settings.ValidityPeriodOption.Companion.toAdvancedOption
+import com.infomaniak.swisstransfer.ui.screen.main.settings.components.SettingOption
 import com.infomaniak.swisstransfer.ui.screen.newtransfer.importfiles.PasswordTransferOption
-import com.infomaniak.swisstransfer.ui.screen.newtransfer.importfiles.TransferAdvancedOptionsEnum
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.launch
@@ -65,7 +65,7 @@ class SettingsViewModel @Inject constructor(
         val passwordOption: PasswordTransferOption,
         val emailLanguage: EmailLanguage,
     ) {
-        fun toAdvancedOptionsList() = listOf<TransferAdvancedOptionsEnum>(
+        fun toAdvancedOptionsList() = listOf<SettingOption>(
             validityPeriod.toAdvancedOption(),
             downloadLimit.toAdvancedOption(),
             passwordOption,
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/SettingsViewModel.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/SettingsViewModel.kt
@@ -25,6 +25,11 @@ import com.infomaniak.multiplatform_swisstransfer.common.models.Theme
 import com.infomaniak.multiplatform_swisstransfer.common.models.ValidityPeriod
 import com.infomaniak.multiplatform_swisstransfer.managers.AppSettingsManager
 import com.infomaniak.swisstransfer.di.IoDispatcher
+import com.infomaniak.swisstransfer.ui.screen.main.settings.DownloadLimitOption.Companion.toAdvancedOption
+import com.infomaniak.swisstransfer.ui.screen.main.settings.EmailLanguageOption.Companion.toAdvancedOption
+import com.infomaniak.swisstransfer.ui.screen.main.settings.ValidityPeriodOption.Companion.toAdvancedOption
+import com.infomaniak.swisstransfer.ui.screen.newtransfer.importfiles.PasswordTransferOption
+import com.infomaniak.swisstransfer.ui.screen.newtransfer.importfiles.TransferAdvancedOptionsEnum
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.launch
@@ -52,5 +57,19 @@ class SettingsViewModel @Inject constructor(
 
     fun setEmailLanguage(emailLanguage: EmailLanguage) = viewModelScope.launch(ioDispatcher) {
         appSettingsManager.setEmailLanguage(emailLanguage)
+    }
+
+    data class AppSettingsData(
+        val validityPeriod: ValidityPeriod,
+        val downloadLimit: DownloadLimit,
+        val isPasswordActivated: Boolean,
+        val emailLanguage: EmailLanguage,
+    ) {
+        fun toAdvancedOptionsList() = listOf<TransferAdvancedOptionsEnum>(
+            validityPeriod.toAdvancedOption(),
+            downloadLimit.toAdvancedOption(),
+            if (isPasswordActivated) PasswordTransferOption.ACTIVATED else PasswordTransferOption.NONE,
+            emailLanguage.toAdvancedOption(),
+        )
     }
 }
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/SettingsViewModel.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/SettingsViewModel.kt
@@ -25,10 +25,6 @@ import com.infomaniak.multiplatform_swisstransfer.common.models.Theme
 import com.infomaniak.multiplatform_swisstransfer.common.models.ValidityPeriod
 import com.infomaniak.multiplatform_swisstransfer.managers.AppSettingsManager
 import com.infomaniak.swisstransfer.di.IoDispatcher
-import com.infomaniak.swisstransfer.ui.screen.main.settings.DownloadLimitOption.Companion.toAdvancedOption
-import com.infomaniak.swisstransfer.ui.screen.main.settings.EmailLanguageOption.Companion.toAdvancedOption
-import com.infomaniak.swisstransfer.ui.screen.main.settings.ValidityPeriodOption.Companion.toAdvancedOption
-import com.infomaniak.swisstransfer.ui.screen.main.settings.components.SettingOption
 import com.infomaniak.swisstransfer.ui.screen.newtransfer.importfiles.PasswordTransferOption
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.CoroutineDispatcher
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -139,18 +187,40 @@ private fun ImportFilesScreen(
...
-                    states = {
-                        listOf(
-                            ValidityPeriodOption.THIRTY,
-                            DownloadLimitOption.ONE,
-                            PasswordTransferOption.NONE,
-                            EmailLanguageOption.GERMAN,
-                        )
-                    },
-                    onClick = {},
+                    modifier = Modifier,
```
tevincent, line 190: Since you're not passing any Modifier, is that necessary ?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -251,6 +327,16 @@ private fun ImportFilesScreenPreview(@PreviewParameter(FileUiListPreviewParamete
...
             currentSessionFilesCount = { 0 },
             selectedTransferType = GetSetCallbacks(get = { TransferType.QR_CODE }, set = {}),
+            advancedOptionsCallbacks = AdvancedOptionsCallbacks(
+                initialValues = SettingsViewModel.AppSettingsData(
+                    validityPeriod = ValidityPeriod.THIRTY,
+                    downloadLimit = DownloadLimit.ONE,
+                    passwordOption = PasswordTransferOption.NONE,
+                    emailLanguage = EmailLanguage.FRENCH,
+                ),
+                advancedOptionsStates = { emptyList() },
```
tevincent, line 337: Preview is not working anymore. We have to add this:

```kotlin
advancedOptionsStates = { listOf(ValidityPeriodOption.THIRTY, DownloadLimitOption.TWO_HUNDRED_FIFTY, PasswordTransferOption.NONE, EmailLanguageOption.FRENCH) },
```

tevincent (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
FabianDevel: of course not, it's oxbow code, good catch

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -98,6 +133,10 @@ private fun ImportFilesScreen(
 ) {
     val context = LocalContext.current
     var showUploadSourceChoiceBottomSheet by rememberSaveable { mutableStateOf(initialShowUploadSourceChoiceBottomSheet) }
+    var showValidityPeriodBottomSheet by rememberSaveable { mutableStateOf(false) }
+    var showDownloadLimitBottomSheet by rememberSaveable { mutableStateOf(false) }
+    var showPasswordOptionBottomSheet by rememberSaveable { mutableStateOf(false) }
+    var showEmailLanguageBottomSheet by rememberSaveable { mutableStateOf(false) }
```
LunarX, line 139: If we had a single state that contains the enum of the currently opened bottom sheet or null otherwise, it could simplify some parts of the code. Have you considered this option?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -142,18 +190,39 @@ private fun ImportFilesScreen(
...
+                closeBottomSheet = { showPasswordOptionBottomSheet = false },
+                initialPosition = advancedOptionsCallbacks.initialValues.passwordOption.ordinal,
+            )
+
+            EmailLanguageBottomSheet(
+                isBottomSheetVisible = { showEmailLanguageBottomSheet },
+                onOptionClicked = { advancedOptionsCallbacks.onAdvancedOptionsValueSelected(it) },
+                closeBottomSheet = { showEmailLanguageBottomSheet = false },
+                initialPosition = advancedOptionsCallbacks.initialValues.emailLanguage.ordinal,
+            )
```
LunarX, line 224: This could probably be split in its own component regrouping all bottom sheets, and I was thinking, if we want to use a mutable state that contains the enum of the currently displayed bottom sheet, then the visibility could be handle with a single `when` that calls the correct BottomSheet component for each case.

If there's no enum mutable state, the group of bottom sheet can still go in its own component anyway

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -57,31 +60,62 @@ private const val TOTAL_FILE_SIZE: Long = 50_000_000_000L
...
+        val advancedOptionsCallbacks = AdvancedOptionsCallbacks(
+            initialValues = SettingsViewModel.AppSettingsData(
+                validityPeriod = safeAppSettings.validityPeriod,
+                downloadLimit = safeAppSettings.downloadLimit,
+                passwordOption = PasswordTransferOption.NONE,
+                emailLanguage = safeAppSettings.emailLanguage,
+            ),
+            advancedOptionsStates = { listOf(validityPeriodState, downloadLimitState, passwordOptionState, emailLanguageState) },
+            onAdvancedOptionsValueSelected = ::onAdvancedOptionsValueSelected,
+        )
```
LunarX, line 98: Right now, everytime the appSettings' state changes, this callback will be recreated because of the "initialValues". This will later recompose all of the screen because it's not a state. To make sure on the correct approach to take, maybe we can talk it out with @sirambd instead of me writing out alternatives I have a few doubts on. Because I don't what know what's the best way to avoid having too much recompositions when only certain parts of the callback have changed.

LunarX: Also initialValues and advancedOptionsStates kind of holds the same data, I feel like there should be a way to only have one does it not? Have you already considered this option?

LunarX: Maybe "initialValues" should be called something else outside of the bottom sheet because outside of the bottom sheet this value simply represents the current setting value.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/TransferAdvancedSettings.kt
```diff
@@ -38,21 +38,21 @@ import com.infomaniak.swisstransfer.ui.images.icons.SpeechBubble
...
 fun TransferAdvancedSettings(
     modifier: Modifier = Modifier,
-    states: () -> List<TransferAdvancedOptionsEnum>,
-    onClick: () -> Unit,
+    states: () -> List<SettingOption?>,
+    onClick: (TransferAdvancedSettingType) -> Unit,
 ) {
     SwissTransferCard(modifier = modifier) {
         TransferAdvancedSettingType.entries.forEach { settingType ->
             val title by remember { derivedStateOf { states()[settingType.ordinal] } }
```
LunarX, line 54: Relying on the list being correctly order feels kind of janky. Is there no alternative?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/TransferAdvancedSetting.kt
```diff
@@ -70,9 +70,9 @@ fun TransferAdvancedSetting(
 }
 
 @Composable
-private fun SettingValue(selectedSetting: () -> TransferAdvancedOptionsEnum) {
+private fun SettingValue(selectedSetting: () -> SettingOption?) {
     Text(
-        text = selectedSetting().title(),
+        text = selectedSetting()?.title?.invoke() ?: "",
```
LunarX, line 75: @sirambd was talking about how we should have whole different empty screen for a split second while waiting for the realm data and only when we received the data can we show the whole screen and at this point all these values would be non-null. What do we do about it?

LunarX (changes requested):
Some first comments

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/TransferAdvancedSetting.kt
FabianDevel: That feels overkill (at least for a beta)

FabianDevel: But a thing that is certain is that it should not be in this PR if we chose to add this

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/TransferAdvancedSettings.kt
FabianDevel: We can have a when on the setting type

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -57,31 +60,62 @@ private const val TOTAL_FILE_SIZE: Long = 50_000_000_000L
...
+        when (option) {
+            is ValidityPeriodOption -> newTransferViewModel.selectTransferValidityPeriod(option)
+            is DownloadLimitOption -> newTransferViewModel.selectTransferDownloadLimit(option)
+            is PasswordTransferOption -> newTransferViewModel.selectTransferPasswordOption(option)
+            is EmailLanguageOption -> newTransferViewModel.selectTransferLanguage(option)
+        }
+    }
+
+    appSettings?.let { safeAppSettings ->
+        newTransferViewModel.initTransferAdvancedOptionsValues(safeAppSettings)
```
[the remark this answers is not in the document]
FabianDevel: Done in the next PR as it would have conflicted

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
FabianDevel: The AppSettings can't be modified during this screen

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/TransferAdvancedSettings.kt
FabianDevel: Fixed together with the use of states instead of initialValues

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
FabianDevel: We merged the usage of initialValues and states

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
FabianDevel: It seems it recomposes the all the options' BottomSheet Scaffold each times now, but this isn't a real performance issue in this case so I'll go with this way

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
FabianDevel: I'll do this in the next PR to avoid conflicts as the password changes

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/NewTransferViewModel.kt
```diff
@@ -129,12 +134,53 @@ class NewTransferViewModel @Inject constructor(
...
+    val selectedTransferType: StateFlow<TransferType> = _selectedTransferType
 
     fun selectTransferType(type: TransferType) {
         _selectedTransferType.value = type
     }
+    //endregion
+
+    //region Transfer Advanced Options
+    private val _selectedValidityPeriodOption = MutableStateFlow<ValidityPeriodOption?>(null)
+    val selectedValidityPeriodOption: StateFlow<ValidityPeriodOption?> = _selectedValidityPeriodOption.asStateFlow()
```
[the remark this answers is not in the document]
FabianDevel: Done in #149

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/SettingsViewModel.kt
```diff
@@ -53,4 +54,11 @@ class SettingsViewModel @Inject constructor(
...
         appSettingsManager.setEmailLanguage(emailLanguage)
     }
+
+    data class AppSettingsData(
+        val validityPeriod: ValidityPeriod,
+        val downloadLimit: DownloadLimit,
+        val passwordOption: PasswordTransferOption,
+        val emailLanguage: EmailLanguage,
+    )
 }
```
LunarX, line 64: Isn't this unused?

PR #149, opened 2024-11-05
feat: Add password popup

FabianDevel:
Depends on #144 

Add the password option that was left out of the #144 PR

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
@@ -72,24 +72,29 @@ fun BasicAlertDialogContent(
     shouldEnableConfirmButton: () -> Boolean = { true },
 ) {
     Column(modifier.padding(Margin.Large)) {
-        Text(
-            stringResource(titleRes),
-            style = SwissTransferTheme.typography.bodyMedium,
-            color = SwissTransferTheme.colors.primaryTextColor,
-        )
-        Spacer(Modifier.height(Margin.Large))
-        Text(
-            stringResource(descriptionRes),
-            style = SwissTransferTheme.typography.bodyRegular,
-            color = SwissTransferTheme.colors.secondaryTextColor,
-        )
-        Spacer(Modifier.height(Margin.Large))
+        TitleAndDescription(titleRes, descriptionRes)
         additionalContent?.let {
             it()
             Spacer(Modifier.height(Margin.Large))
         }
         ActionButtons(onDismissRequest, onConfirmation, shouldEnableConfirmButton)
     }
+}
+
+@Composable
+private fun TitleAndDescription(titleRes: Int, descriptionRes: Int) {
+    Text(
+        text = stringResource(titleRes),
+        style = SwissTransferTheme.typography.bodyMedium,
+        color = SwissTransferTheme.colors.primaryTextColor,
+    )
+    Spacer(Modifier.height(Margin.Large))
+    Text(
+        text = stringResource(descriptionRes),
+        style = SwissTransferTheme.typography.bodyRegular,
+        color = SwissTransferTheme.colors.secondaryTextColor,
+    )
+    Spacer(Modifier.height(Margin.Large))
 }
 
 @Composable
@@ -127,4 +132,3 @@ private fun PreviewAlertDialog() {
         }
     }
 }
-
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
@@ -20,14 +20,13 @@ package com.infomaniak.swisstransfer.ui.components
 import androidx.annotation.StringRes
 import androidx.compose.foundation.layout.*
 import androidx.compose.foundation.shape.RoundedCornerShape
-import androidx.compose.material3.BasicAlertDialog
-import androidx.compose.material3.Card
-import androidx.compose.material3.ExperimentalMaterial3Api
-import androidx.compose.material3.Text
+import androidx.compose.material3.*
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.stringResource
-import androidx.compose.ui.unit.dp
+import androidx.compose.ui.tooling.preview.Preview
+import com.infomaniak.swisstransfer.R
 import com.infomaniak.swisstransfer.ui.theme.Margin
 import com.infomaniak.swisstransfer.ui.theme.SwissTransferTheme
 
@@ -47,8 +46,8 @@ fun SwissTransferAlertDialog(
             .wrapContentHeight()
             .fillMaxWidth(),
     ) {
-        Card(shape = RoundedCornerShape(28.dp)) {
-            BasicAlertDialogContent(modifier, titleRes, descriptionRes, content)
+        Card(shape = RoundedCornerShape(Margin.Medium)) {
+            BasicAlertDialogContent(modifier, titleRes, descriptionRes, content, onDismissRequest)
         }
     }
 }
@@ -59,6 +58,7 @@ fun BasicAlertDialogContent(
     @StringRes titleRes: Int,
     @StringRes descriptionRes: Int,
     additionalContent: @Composable (ColumnScope.() -> Unit)? = null,
+    onDismissRequest: () -> Unit,
 ) {
     Column(modifier.padding(Margin.Large)) {
         Text(
@@ -74,6 +74,42 @@ fun BasicAlertDialogContent(
         )
         Spacer(Modifier.height(Margin.Large))
         additionalContent?.let { it() }
+        ActionButtons(onDismissRequest, {})
+
+    }
+}
+
+@Composable
+private fun ActionButtons(onDismissRequest: () -> Unit, onConfirmation: () -> Unit) {
+    Row(
+        modifier = Modifier.fillMaxWidth(),
+        horizontalArrangement = Arrangement.End,
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        SmallButton(
+            style = ButtonType.TERTIARY,
+            titleRes = R.string.buttonCancel,
+            onClick = { onDismissRequest() },
+        )
+        Spacer(Modifier.width(Margin.Micro))
+        SmallButton(
+            titleRes = R.string.buttonCancel,
+            onClick = { onConfirmation() },
+        )
+    }
+}
+
+@Preview
+@Composable
+private fun PreviewAlertDialog() {
+    SwissTransferTheme {
+        Surface {
+            SwissTransferAlertDialog(
+                titleRes = R.string.uploadSuccessLinkTitle,
+                descriptionRes = R.string.uploadSuccessLinkDescription,
+                onDismissRequest = {}
+            ) { }
+        }
     }
 }
 
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
@@ -73,6 +73,7 @@ private fun BasicAlertDialogContent(
 ) {
     Column(modifier.padding(Margin.Large)) {
         TitleAndDescription(titleRes, descriptionRes)
+        Spacer(Modifier.height(Margin.Large))
         additionalContent?.let {
             it()
             Spacer(Modifier.height(Margin.Large))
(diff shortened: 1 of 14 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/NewTransferViewModel.kt
```diff
@@ -179,6 +183,15 @@ class NewTransferViewModel @Inject constructor(
     }
     //endregion
 
+    //region Password
+    var transferPassword by mutableStateOf("")
+
+    val isPasswordValid by derivedStateOf {
+        val trimmedPassword = transferPassword.trim()
```
tevincent, line 190: Should we not allow spaces on the password field ? I'm afraid this trim might cause issues.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/NewTransferViewModel.kt
```diff
@@ -179,6 +183,15 @@ class NewTransferViewModel @Inject constructor(
...
+        trimmedPassword.length in 6..25
```
tevincent, line 191: Should we put these magic number as constants ?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/PasswordOptionAlertDialog.kt
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    if (showPasswordOptionAlert()) {
+        SwissTransferAlertDialog(
+            titleRes = R.string.settingsOptionPassword,
+            descriptionRes = R.string.settingsPasswordDescription,
+            onDismissRequest = ::onDismissRequest,
+            onConfirmation = ::onConfirmButtonClicked,
+            shouldEnableConfirmButton = { if (isPasswordActivated) isPasswordValid() else true },
+        ) {
```
tevincent, line 80: When clicking on the password field, the `ImeAction.DONE` is missing ? We might need to add this to the SwissTransferTextField -> OutlineTextField :

```kotlin
...
keyboardOptions = KeyboardOptions(
  keyboardType = if (isPassword) KeyboardType.Password else keyboardType,
  autoCorrectEnabled = true,
  imeAction = ImeAction.Done,
),
keyboardActions = KeyboardActions(
  onDone = { /* your callback here */},
),
 ...
```

Of course, we need to make this configurable.

tevincent (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/NewTransferViewModel.kt
FabianDevel: That sound like a bad practice from their side no ?

FabianDevel: We'll talk about entirely blocking users from adding space on team meeting

tevincent: DO we want to keep the password as is ? The user want a specific password and modifying it is dangerous ?

tevincent (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/NewTransferViewModel.kt
LunarX, line 1: You can use this strategy to interact with the StateFlow saved in the savedStateHandle: https://developer.android.com/topic/libraries/architecture/viewmodel/viewmodel-savedstate#savedstate-stateflow

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -197,6 +229,41 @@ private fun ColumnScope.EmailAddressesTextFields(selectedTransferType: () -> Tra
     }
 }
 
+@Composable
+private fun AdvancedOptions(
```
LunarX, line 233: "AdvancedOptions" already kind of describes the actual options always present at the bottom of the screen. Maybe this component should have something relating to "BottomSheets" in its name?

LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/NewTransferViewModel.kt
```diff
@@ -129,12 +143,110 @@ class NewTransferViewModel @Inject constructor(
...
+                selectTransferDownloadLimit(it.downloadLimit.toAdvancedOption())
+                selectTransferLanguage(it.emailLanguage.toAdvancedOption())
+            } ?: run {
+                SentryLog.e(TAG, "AppSettings is null in ImportFilesScreen, it should not happened")
+                selectTransferValidityPeriod(ValidityPeriodOption.THIRTY)
+                selectTransferDownloadLimit(DownloadLimitOption.TWO_HUNDRED_FIFTY)
+                selectTransferLanguage(EmailLanguageOption.FRENCH)
+            }
+        }
+    }
```
LunarX, line 209: I was discussing this with @sirambd, but maybe there's an even better solution than using LaunchedEffects or ViewMode.init(). I feel like it might be applicable here maybe? This is still kinda uncertain maybe?

Anyway, the idea is to prefer cold flows that kick off their work when you start collecting them via collectAsStateWithLifecycle()

Source: https://proandroiddev.com/loading-initial-data-in-launchedeffect-vs-viewmodel-f1747c20ce62

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
FabianDevel: Except the password one is an alert dialog, not a bottomsheet

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/NewTransferViewModel.kt
FabianDevel: It's interesting, we could do this later

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -284,6 +246,41 @@ private fun ColumnScope.EmailAddressesTextFields(selectedTransferType: () -> Tra
...
+    selectedTransferType: () -> TransferAdvancedSettingType?,
```
KevinBoulongne, line 251: Shouldn't this be called `selectedSetting` ?

Something about setting, and not about transfer.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
KevinBoulongne: The options themselves are named `TransferAdvancedSettings()`
So this function should be about `Settings` and not `Options.`

For the full name, maybe something like `AdvancedSettingsDisplay` ?

I think we could also remove the `Transfer` part in `TransferAdvancedSettings()`, it's kinda redundant.

_(Instead, we could rename `Settings` into `Options` if there is too much places where `Options` is already used.)_

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
```diff
@@ -0,0 +1,134 @@
...
+    modifier: Modifier = Modifier,
+    @StringRes titleRes: Int,
+    @StringRes descriptionRes: Int,
+    onDismiss: () -> Unit,
+    onConfirmation: () -> Unit,
+    shouldEnableConfirmButton: () -> Boolean = { true },
+    content: @Composable (ColumnScope.() -> Unit)? = null,
+) {
+    BasicAlertDialog(
+        onDismissRequest = onDismiss,
```
KevinBoulongne, line 45: Shouldn't this one be renamed `onDismiss` too ?

KevinBoulongne: … this one is in the Android SDK 🤣 

Can you do a PR on the AOSP repository to rename it, please ? 😁

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
```diff
@@ -0,0 +1,134 @@
...
+        style = SwissTransferTheme.typography.bodyMedium,
+        color = SwissTransferTheme.colors.primaryTextColor,
+    )
+    Spacer(Modifier.height(Margin.Large))
+    Text(
+        text = stringResource(descriptionRes),
+        style = SwissTransferTheme.typography.bodyRegular,
+        color = SwissTransferTheme.colors.secondaryTextColor,
+    )
+    Spacer(Modifier.height(Margin.Large))
```
KevinBoulongne, line 97: This Spacer is kinda hidden in here.

Shouldn't we move it to `BasicAlertDialogContent` ?

The concept of using `Spacer` instead of `Modifier.padding` is to more easily see the margins between components.
So I feel we should avoid putting Spacers inside components if they are on the edges of the component _(`top` and `bot` edges for Columns, `start` and `end` edges for Rows)_.

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTextField.kt
```diff
@@ -72,6 +76,12 @@ fun SwissTransferTextField(
         unfocusedSupportingTextColor = SwissTransferTheme.colors.tertiaryTextColor,
     )
 
+    @Composable
+    fun getSupportingText(): (@Composable () -> Unit)? {
+        val displayText = if (text.isEmpty()) supportingText else errorMessage() ?: supportingText
```
KevinBoulongne, line 81: Could you add parenthesis or brackets in this `if / else / ?:` combination please ?

I feel that this way of writing the condition is error prone.
I don't know exactly which `displayText` will be chosen.

KevinBoulongne:
``` kotlin
val displayText = (if (text.isEmpty()) supportingText else errorMessage()) ?: supportingText
```
or
``` kotlin
val displayText = if (text.isEmpty()) {
    supportingText
} else {
    errorMessage() ?: supportingText
}
```

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
FabianDevel: But it's about the transfer, not the settings.
Setting make it looks like it belongs to the application settings that can be find in the setting screen

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
FabianDevel: We should probably go with options to avoid confusion with the general settings

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
FabianDevel: Yeah sure 😂

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
FabianDevel: I like the idea that a component takes care of everything to be exactly like we want it to be. And as there is no cases where we won't need a margin after the description, it feels right to me that the component has its spacer.
But I also understand we could want to see all the spacer in the "parent" component to separate its children

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
KevinBoulongne: I'm gonna handle the renaming in another PR.

LunarX (approved):
The renaming I proposed and the savedStateHandle interaction with flows is approved

KevinBoulongne (approved):

PR #160, opened 2024-11-08
Update share and copy link button's design

LunarX:
The old design with 3 buttons on top of one another has been scrapped for a design with two bigger buttons but side by side this time around

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/upload/UploadSuccessQrScreen.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/upload/UploadSuccessQrScreen.kt
@@ -58,7 +58,7 @@ fun UploadSuccessQrScreen(transferType: TransferType, transferUrl: String, close
         },
     ) {
         Column {
-            ScreenContent(transferType, transferUrl)
+            SuccessMessage(transferType, transferUrl)
 
             ShareAndCopyButtons(
                 modifier = Modifier.padding(bottom = Margin.Medium, top = Margin.Mini),
@@ -70,10 +70,7 @@ fun UploadSuccessQrScreen(transferType: TransferType, transferUrl: String, close
 }
 
 @Composable
-private fun ColumnScope.ScreenContent(
-    transferType: TransferType,
-    transferUrl: String
-) {
+private fun ColumnScope.SuccessMessage(transferType: TransferType, transferUrl: String) {
     Column(
         modifier = Modifier
             .fillMaxWidth()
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/upload/UploadSuccessQrScreen.kt
```diff
@@ -27,114 +26,91 @@ import androidx.compose.material3.Surface
...
-            context.copyText(
-                text = transferLink,
-                showSnackbar = { scope.launch { snackbarHostState.showSnackbar(it) } },
-            )
-        },
-    )
+        ShareAndCopyButtons(
+            modifier = Modifier
+                .padding(horizontal = Margin.Medium)
+                .padding(bottom = Margin.Medium, top = Margin.Mini),
```
FabianDevel, line 109: in the case here where we already have to dissociate the padding, do we really want to double the padding() call ? 
I don't really know what I would prefer as both versions have their problem

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/upload/components/ShareAndCopyButtons.kt
```diff
@@ -0,0 +1,111 @@
...
+import com.infomaniak.swisstransfer.ui.theme.SwissTransferTheme
+import com.infomaniak.swisstransfer.ui.utils.PreviewLightAndDark
+import com.infomaniak.swisstransfer.ui.utils.copyText
+import com.infomaniak.swisstransfer.ui.utils.shareText
+import kotlinx.coroutines.launch
+
+@Composable
+fun ColumnScope.ShareAndCopyButtons(modifier: Modifier = Modifier, transferLink: String, snackbarHostState: SnackbarHostState) {
+    val scope = rememberCoroutineScope()
+    val context = LocalContext.current
```
FabianDevel, line 48: This variable doesn't seem necessary.
 `.current` is just a direct access as it is scoped right?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/upload/components/ShareAndCopyButtons.kt
```diff
@@ -0,0 +1,111 @@
...
+            verticalArrangement = Arrangement.Center,
+        ) {
+            Icon(icon, contentDescription = null, modifier = Modifier.size(Dimens.SmallIconSize))
+            Spacer(modifier = Modifier.height(Margin.Micro))
+            Text(stringResource(textRes))
+        }
+    }
+}
+
+
```
FabianDevel, line 96:
```suggestion

```

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/upload/UploadSuccessQrScreen.kt
LunarX: Chaining the paddings is pretty standard I think. This solution still has the benefit of conveying the idea that start and end are linked together which is really what I try to enforce on that line

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/upload/components/ShareAndCopyButtons.kt
```diff
@@ -0,0 +1,111 @@
...
+            onClick = {
+                context.copyText(
+                    text = transferLink,
+                    showSnackbar = { scope.launch { snackbarHostState.showSnackbar(it) } },
+                )
+            },
+        )
+    }
+}
+@Composable
```
tevincent, line 76:
```suggestion
}

@Composable
```

tevincent (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/upload/components/ShareAndCopyButtons.kt
LunarX: Yeah, unfortunately, here it is indeed needed because the `onClick()` callbacks are not composable functions but `current` is a composable function. This prevents us from calling `.current` inside the `onClick()` directly. Otherwise, yes definitely

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/upload/UploadSuccessQrScreen.kt
LunarX: But I realized that the horizontal padding needs to go inside the component actually, so the paddings are not chained here anymore

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/upload/UploadSuccessQrScreen.kt
```diff
@@ -27,114 +26,95 @@ import androidx.compose.material3.Surface
...
+        }
+    }
 }
 
 @Composable
-private fun Content(context: Context, snackbarHostState: SnackbarHostState, transferType: TransferType, transferUrl: String) {
-
-    val scope = rememberCoroutineScope()
-
+private fun ColumnScope.ScreenContent(
```
FabianDevel, line 73: This screen content composable seems overkill, also I don't really like this because the ShareAndCopyButton are screen content too, but aren't inside the composable named this way

FabianDevel (changes requested):

FabianDevel (approved):

PR #164, opened 2024-11-11
refactor: Rename Transfer Options related code

KevinBoulongne:


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/TransferOption.kt
```diff
@@ -38,21 +38,21 @@ import com.infomaniak.swisstransfer.ui.theme.SwissTransferTheme
 import com.infomaniak.swisstransfer.ui.utils.PreviewLightAndDark
 
 @Composable
-fun TransferAdvancedSetting(
-    settingType: TransferAdvancedSettingType,
+fun TransferOption(
```
FabianDevel, line 41: Shouldn't this class be `TransferOptionType` ? As it represent one item of  `TransferOptionsTypes`

FabianDevel (changes requested):

FabianDevel (approved):

PR #165, opened 2024-11-12
feat: Make Sent screen use the ScreenWrapper

KevinBoulongne:
Depends on #173

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItemList.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItemList.kt
@@ -27,6 +27,7 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.PreviewParameter
 import com.infomaniak.multiplatform_swisstransfer.common.interfaces.ui.TransferUi
+import com.infomaniak.multiplatform_swisstransfer.common.models.TransferDirection
 import com.infomaniak.swisstransfer.R
 import com.infomaniak.swisstransfer.ui.previewparameter.TransferUiListPreviewParameter
 import com.infomaniak.swisstransfer.ui.theme.Margin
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItemList.kt
```diff
@@ -36,12 +37,14 @@ import com.infomaniak.swisstransfer.ui.utils.PreviewLightAndDark
...
 fun TransferItemList(
     modifier: Modifier = Modifier,
+    direction: TransferDirection,
     transfers: List<TransferUi>,
     getSelectedTransferUuid: () -> String?,
     onClick: (TransferUi) -> Unit,
 ) {
 
     val selectedTransferUuid = getSelectedTransferUuid()
+    val titleRes = if (direction == TransferDirection.SENT) R.string.sentFilesTitle else R.string.receivedFilesTitle
```
FabianDevel, line 47: Maybe a `when` would be more suitable as it's an enum, so we don't risk a potential bad value in the hypothetical case we add smt in the enum

app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItemList.kt
```diff
@@ -51,7 +54,7 @@ fun TransferItemList(
 
         item {
             Text(
-                text = stringResource(R.string.receivedFilesTitle),
+                text = stringResource(titleRes),
                 style = SwissTransferTheme.typography.h1,
             )
```
FabianDevel, line 59: This could maybe be onlined

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transfers/TransfersScreenWrapper.kt
```diff
@@ -28,29 +28,38 @@ import androidx.compose.material3.adaptive.navigation.ThreePaneScaffoldNavigator
...
+private fun ListPane(direction: TransferDirection, navigator: ThreePaneScaffoldNavigator<String>) {
+    when (direction) {
+        TransferDirection.SENT -> SentScreen(
+            navigateToDetails = { transferUuid -> navigator.navigateTo(ListDetailPaneScaffoldRole.Detail, transferUuid) },
+            getSelectedTransferUuid = { navigator.currentDestination?.content },
+        )
+        TransferDirection.RECEIVED -> ReceivedScreen(
+            navigateToDetails = { transferUuid -> navigator.navigateTo(ListDetailPaneScaffoldRole.Detail, transferUuid) },
+            getSelectedTransferUuid = { navigator.currentDestination?.content },
+        )
```
FabianDevel, line 61: these two function could be factorized

FabianDevel (changes requested):
I wonder if it would not be easier to simply have a single screen for both received and sent that takes the particularities of each in callbacks passed from the wrapper, as this wrapper already check the direction

tevincent (approved):

sirambd (approved):

FabianDevel (approved):

PR #189, opened 2024-11-14
fix: Use UiState pattern to load TransferDetailsScreen

tevincent:


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/components/TransferInfo.kt
```diff
@@ -49,8 +49,8 @@ import com.infomaniak.swisstransfer.ui.utils.PreviewLightAndDark
 @Composable
 fun TransferInfo(transfer: TransferUi) {
 
-    val filesCount = transfer.files.count()
-    val downloadedCount by remember { derivedStateOf { transfer.downloadLimit - transfer.downloadLeft } }
+    val filesCount  by remember(transfer) { derivedStateOf { transfer.files.count() } }
```
LunarX, line 52: You should not use a derivedStateOf if there's no significant optimization. 

In other words, if you have an event "A" and you have a value you want to compute "B" which value depends on A. If A changing to a different value implies that B will change to a different value as well, then you don't want to use a derivedStateOf. You want to use a derivedStateOf when A changing to a different value can lead to B staying the same value.

LunarX (changes requested):

KevinBoulongne (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/TransferDetailsViewModel.kt
```diff
@@ -54,4 +60,16 @@ class TransferDetailsViewModel @Inject constructor(
...
+    sealed class TransferDetailsUiState {
+
+        @Immutable
+        data class Success(val transfer: TransferUi) : TransferDetailsUiState()
+
+        @Immutable
+        data object Loading : TransferDetailsUiState()
+
+        @Immutable
+        data object Error : TransferDetailsUiState() //TODO Handle error case
```
tevincent, line 73: Add `Delete` state.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/components/TransferInfo.kt
KevinBoulongne: This conversation is marked as resolved, but the issue seems to still be here.

Is it on purpose ?

tevincent: After a discussion with @sirambd and @LunarX , we're keeping this as is.

sirambd (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/TransferDetailsScreen.kt
```diff
@@ -298,7 +299,7 @@ private fun Preview(@PreviewParameter(TransferUiListPreviewParameter::class) tra
                 transferUrl = "",
                 direction = TransferDirection.SENT,
                 navigateBack = null,
-                getTransfer = { transfers.first() },
+                getTransfer = { Success(transfers.first()).transfer },
```
KevinBoulongne, line 302: Why do we create a `data class` only to get the Transfer that we already have ?

KevinBoulongne (changes requested):

LunarX (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/TransferDetailsScreen.kt
LunarX: How I see things is that the "data class" here is first used to convey the state of the ui. 

Then, when the state is "success", we also want to know what transfer to display in the UI and can use this `Success` data class to pass it down

LunarX: I like to think about it like having an enum except that you can have different signatures for each value of the enum. This way you can have only relevant data inside each value of the "enum".

But don't take my words for it, I'd still be curious about the other's answer

KevinBoulongne (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/TransferDetailsScreen.kt
LunarX: Nevermind you didn't ask bout the UiState

PR #190, opened 2024-11-14
fix: Put the correct mimeType when importing files

tevincent:
Before:

https://github.com/user-attachments/assets/14f7ffb0-d0c9-4b50-bbc6-49ca22a9f621

After:

https://github.com/user-attachments/assets/a50175ed-1641-4766-b949-77ae9cd5eb98

### DIFF
--- a/FileTypes/src/main/java/com/infomaniak/library/filetypes/FileType.kt
+++ b/FileTypes/src/main/java/com/infomaniak/library/filetypes/FileType.kt
@@ -46,14 +46,15 @@ enum class FileType(val icon: ImageVector, private val colorLight: Color, privat
     }
 
     companion object {
-        fun getMimeTypeFromFileName(fileName: String): String? {
+        //TODO Move this method into Core2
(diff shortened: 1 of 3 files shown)


FileTypes/src/main/java/com/infomaniak/library/filetypes/FileType.kt
```diff
@@ -46,9 +46,14 @@ enum class FileType(val icon: ImageVector, private val colorLight: Color, privat
     }
 
     companion object {
+        fun getMimeTypeFromFileName(fileName: String): String? {
+            return fileName.extractExtension()?.let { extension ->
+                MimeTypeMap.getSingleton().getMimeTypeFromExtension(extension)
+            }
+        }
+
```
LunarX, line 54: Maybe rename it to `guess` instead of `get` and add a TODO to move it into Core2. When we have a correct structure in Core2 we should move this code in the common part of Core2 maybe as it's kind of independent of FileType

LunarX (approved):

PR #202, opened 2024-11-19
Animate highlighted text

LunarX:
When the higlighter highlights some text, it animates the highlight gradually from left to right


TommyDL-Infomaniak (approved):

NicolasBourdin88 (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/HighlightedText.kt
```diff
@@ -86,13 +105,19 @@ private fun List<Rect>.transformForHighlightedStyle(
...
-        addPath(boundingBox.transformForHighlightedStyle(verticalPadding, horizontalPadding, angleDegrees))
+        addPath(boundingBox.transformForHighlightedStyle(verticalPadding, horizontalPadding, angleDegrees, progress))
     }
 }
 
-private fun Rect.transformForHighlightedStyle(verticalPadding: Float, horizontalPadding: Float, angleDegrees: Float): Path {
+private fun Rect.transformForHighlightedStyle(
+    verticalPadding: Float,
+    horizontalPadding: Float,
+    angleDegrees: Float,
```
tevincent, line 118: Is there a reason to pass `angleDegrees` as a `Float` and then convert it to a `Double` at the end ? Just convert your constant to a `Double` and you're good to go.

```kotlin
private const val ROTATION_ANGLE_DEGREE = -3.0
```

tevincent (changes requested):

tevincent (approved):

PR #214, opened 2024-11-27
feat: Use `SwipeToDismissBoxState` to be able to delete Transfers

KevinBoulongne:
Depends on #211


app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItemList.kt
```diff
@@ -63,11 +67,17 @@ fun TransferItemList(
             contentType = { getTransfers()[it] },
             itemContent = {
                 val transfer = getTransfers()[it]
-                TransferItem(
-                    transfer = transfer,
-                    isSelected = { selectedTransferUuid == transfer.uuid },
-                    onClick = { onClick(transfer) },
-                )
+                SwipeToDismissComponent(
```
tevincent, line 70: Is it better to directly call `CustomShapes.SMALL` instead of creating another `val` ? It's not costly since `CustomShapes.SMALL` is a constant and we don't have to look way on the top to know what is `itemShape`.

app/src/main/java/com/infomaniak/swisstransfer/ui/images/icons/Bin.kt
```diff
@@ -42,88 +42,82 @@ val AppIcons.Bin: ImageVector
 
         _bin = Builder(
             name = "Bin",
-            defaultWidth = 16.0.dp,
-            defaultHeight = 16.0.dp,
-            viewportWidth = 16.0f,
-            viewportHeight = 16.0f,
+            defaultWidth = 24.0.dp,
```
tevincent, line 45: Capture d’écran 2024-11-29 à 16 42 14

Is it me or the bin is blurry when zoomed in ? I might need glasses.

tevincent (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItemList.kt
KevinBoulongne: The shape is used twice, and it needs to be the same.

app/src/main/java/com/infomaniak/swisstransfer/ui/images/icons/Bin.kt
KevinBoulongne: It seems that it depends of the device.
I have the same feeling as you on a Freeform.

But on Pixel 7, it's perfectly smooth.

tevincent (approved):

PR #219, opened 2024-11-28
feat: Auto-delete old expired Transfers

KevinBoulongne:
Depends on #229

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/MainApplication.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/MainApplication.kt
@@ -20,9 +20,13 @@ package com.infomaniak.swisstransfer.ui
 import android.app.Application
 import androidx.hilt.work.HiltWorkerFactory
 import androidx.work.Configuration
+import com.infomaniak.multiplatform_swisstransfer.common.interfaces.ui.TransferUi
+import com.infomaniak.multiplatform_swisstransfer.common.models.TransferDirection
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/MainApplication.kt
```diff
@@ -65,4 +73,17 @@ class MainApplication : Application(), Configuration.Provider {
             }
         }
     }
+
+    private suspend fun collectTransfers() {
+        transferManager.getTransfers(TransferDirection.SENT).collect(::deleteExpiredTransfers)
+        transferManager.getTransfers(TransferDirection.RECEIVED).collect(::deleteExpiredTransfers)
```
tevincent, line 79: Deleting expired transfers might be on KMP ? That way, we avoid doing the work on both platform ?

tevincent (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/MainApplication.kt
KevinBoulongne: Yes totally, the logic to detect an old expired Transfer could be shared.

KevinBoulongne: It's been moved.

tevincent (approved):

PR #220, opened 2024-11-28
refactor: Polish import screen

FabianDevel:


LunarX (changes requested):
First comments, the rest will come later

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTextField.kt
```diff
@@ -119,8 +113,8 @@ fun SwissTransferTextField(
             imeAction = imeAction,
         ),
         keyboardActions = keyboardActions,
-        isError = errorMessage() != null && text.isNotEmpty(),
-        supportingText = getSupportingText(),
+        isError = shouldDisplayError,
```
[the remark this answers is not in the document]
FabianDevel: Yes finally, it seems it would be better to let it as it is
Simplifiyng its usage make it less versatile

app/src/main/java/com/infomaniak/swisstransfer/ui/components/DoubleButtonCombo.kt
```diff
@@ -71,7 +71,7 @@ private fun VerticallyStackedButtons(
         bottomButton(
             Modifier
                 .fillMaxWidth()
-                .padding(start = Margin.Medium, end = Margin.Medium, bottom = Margin.Large),
+                .padding(start = Margin.Medium, end = Margin.Medium, bottom = Margin.Small),
```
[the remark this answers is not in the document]
LunarX: I just read that comment again but if it's only used in this file, we can settle for a private const instead

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -189,39 +171,66 @@ private fun ImportFilesScreen(
...
+        modifier = Modifier.padding(Margin.Medium),
+        files = files,
+        humanReadableSize = { humanReadableSize },
+        pickFiles = ::pickFiles,
+        removeFileByUid = removeFileByUid,
+    )
+}
+
+@Composable
+private fun ColumnScope.ImportTextFields(
```
LunarX, line 226: I don't think you need the `ColumnScope`

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -230,28 +239,46 @@ private fun ColumnScope.ImportTextFields(transferMessage: GetSetCallbacks<String
 }
 
 @Composable
-private fun ColumnScope.EmailAddressesTextFields(selectedTransferType: () -> TransferTypeUi) {
+private fun ColumnScope.EmailAddressesTextFields(modifier: Modifier, selectedTransferType: () -> TransferTypeUi) {
```
LunarX, line 242: I don't think you need the `ColumnScope`

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -189,39 +171,66 @@ private fun ImportFilesScreen(
         content = {
             Column(
                 modifier = Modifier
-                    .padding(horizontal = Margin.Medium, vertical = Margin.Large)
+                    .widthIn(max = Dimens.MaxSinglePaneScreenWidth)
```
LunarX, line 174: I think the max width should go somewhere higher if possible so it's applied to all of the screens of this activity

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -189,39 +171,66 @@ private fun ImportFilesScreen(
...
+        contract = ActivityResultContracts.OpenMultipleDocuments(),
+        onResult = addFiles,
+    )
 
+    fun pickFiles() {
+        shouldShowInitialFilePick = false
+        filePickerLauncher.launch(arrayOf("*/*"))
+    }
+
+    LaunchedEffect(Unit) { if (shouldShowInitialFilePick) pickFiles() }
```
LunarX, line 213: Having this logic be defined here somewhat feels unnatural yet this is the only place where it is needed so I'm not sure what to think of it

LunarX (changes requested):
A few more comments, the rest will once again come later

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
FabianDevel: It is needed by the animated visibilty to have a pretty anim

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
FabianDevel: same

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
FabianDevel: We decided to keep it here to avoid overhead code

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/PasswordOptionAlertDialog.kt
```diff
@@ -75,9 +72,10 @@ fun PasswordOptionAlertDialog(
         descriptionRes = R.string.settingsPasswordDescription,
         onDismiss = ::onDismiss,
         onConfirmation = ::onConfirmButtonClicked,
-        shouldEnableConfirmButton = { if (isPasswordActivated) isPasswordValid() else true },
+        shouldEnableConfirmButton = { !isPasswordActivated || isPasswordValid() },
```
LunarX, line 75: I feel like we often we have properties called `isEnabled` in google's components. Do you think we could do closer to the usual naming? Maybe something like `isConfirmButtonEnabled`?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/PasswordOptionAlertDialog.kt
```diff
@@ -102,13 +100,17 @@ private fun ColumnScope.AnimatedPasswordInput(
...
+            modifier = Modifier.fillMaxWidth(),
             label = stringResource(R.string.settingsOptionPassword),
             isPassword = true,
             initialValue = password.get(),
             imeAction = ImeAction.Done,
-            errorMessage = { if (isPasswordValid()) null else stringResource(R.string.errorTransferPasswordLength) },
+            isError = !isPasswordValid(),
+            supportingText = { isError ->
+                val errorText = if (isError) stringResource(R.string.errorTransferPasswordLength) else ""
+                Text(text = errorText, minLines = 2)
```
LunarX, line 112: When going the "alpha route" minLines won't be needed anymore, I just add this here but you might have already modified it

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/TransferTypeButtons.kt
```diff
@@ -42,7 +45,9 @@ import com.infomaniak.swisstransfer.ui.utils.PreviewLightAndDark
 @Composable
 fun TransferTypeButtons(transferType: GetSetCallbacks<TransferTypeUi>) {
     Row(
-        modifier = Modifier.horizontalScroll(rememberScrollState()),
+        modifier = Modifier
+            .horizontalScroll(rememberScrollState())
+            .windowInsetsPadding(insets = WindowInsets(left = Margin.Medium, top = 0.dp, right = Margin.Medium, bottom = 0.dp)),
```
LunarX, line 50: Why use a `windowInsetsPadding` over a `padding`?

LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/TransferTypeButtons.kt
FabianDevel: I was convinced that adding padding wasn't working because the padding wasn't scrollable, I guess I was just running old version of code, it works indeed

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
FabianDevel: I moved it to StockyBottomButtonScafflod as all screen using it should be constrained like this

LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/BottomStickyButtonScaffold.kt
```diff
@@ -40,9 +43,17 @@ fun BottomStickyButtonScaffold(
...
                 .fillMaxWidth()
-                .padding(contentPaddings),
+                .padding(contentPaddings)
+                .padding(vertical = Margin.Small),
+            horizontalAlignment = Alignment.CenterHorizontally,
         ) {
-            Box(modifier = Modifier.weight(1.0f), content = content)
+            Box(
+                modifier = Modifier
+                    .padding(bottom = Margin.Small)
```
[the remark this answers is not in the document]
FabianDevel: Yes I wanted to set also the padding at the top of the screen to keep consistancy between the top of the content and the bottom of the screen
Like it should have the same padding between the topBar/content, content/stickyButton and stickyButton/Bottom of screen

I can indeed make a const for the padding

LunarX (changes requested):
We also need to make sure the bottom sheet scaffold that also uses the `DoubleButtonCombo` uses the correct padding

app/src/main/java/com/infomaniak/swisstransfer/ui/theme/Dimens.kt
```diff
@@ -20,13 +20,15 @@ package com.infomaniak.swisstransfer.ui.theme
...
     val SettingVerticalMargin = Margin.Small
     val DescriptionWidth = 300.dp
     val LargeButtonHeight = 56.dp
-    val DoubleButtonMaxWidth = 800.dp
+    val DoubleButtonMaxWidth = MaxSinglePaneScreenWidth
     val SingleButtonMaxWidth = DoubleButtonMaxWidth / 2
     val SmallIconSize = 16.dp
     val IconSize = 24.dp
     val BorderWidth = 1.dp
+    val ButtonComboVerticalPadding = Margin.Small
```
LunarX, line 33: The same way, I'd not put the margins directly inside `DoubleButtonCombo` so they are not forced everywhere this component is used, I also wouldn't have a dimen that's called `ButtonComboVerticalPadding` to not force it everytime a `DoubleButtonCombo` is used. Else this would directly be defined inside `DoubleButtonCombo`.

Wouldn't `ScaffoldButtonComboVerticalPadding` be better, this way it limits the scope to only scaffolds but not like all of the possible usages?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -49,11 +48,8 @@ import com.infomaniak.swisstransfer.ui.screen.newtransfer.importfiles.components
 import com.infomaniak.swisstransfer.ui.theme.Margin
 import com.infomaniak.swisstransfer.ui.theme.SwissTransferTheme
 import com.infomaniak.swisstransfer.ui.utils.GetSetCallbacks
-import com.infomaniak.swisstransfer.ui.utils.HumanReadableSizeUtils.getHumanReadableSize
 import com.infomaniak.swisstransfer.ui.utils.PreviewAllWindows
 
-private const val TOTAL_FILE_SIZE: Long = 50_000_000_000L
-
 @Composable
 fun ImportFilesScreen(
```
KevinBoulongne, line 54: We should talk about the paddings in this file.

KevinBoulongne (changes requested):

LunarX (approved):

KevinBoulongne (approved):

PR #221, opened 2024-11-28
feat: Enable user to see their password

FabianDevel:


gradle/libs.versions.toml
```diff
@@ -20,7 +20,7 @@ qrose = "1.0.1"
 recaptcha = "18.6.1"
 sentry = "4.12.0"
 serialization = "1.7.3"
-swisstransfer = "0.9.0"
+swisstransfer = "0.9.2"
```
KevinBoulongne, line 23: We need to wait to KMP 0.9.2 to be available before merging this PR.

KevinBoulongne (changes requested):

gradle/libs.versions.toml
KevinBoulongne, line 23: Maybe we should do this bump in its own PR, before this one ?

KevinBoulongne (changes requested):

sirambd (approved):

tevincent (approved):

gradle/libs.versions.toml
KevinBoulongne: image

KevinBoulongne (approved):

PR #222, opened 2024-11-29
feat: Onboarding

LunarX:
This PR still lacks the correct pager indicator, but for now I've put a quick placeholder code and I will code and style the indicator in the following PR.

This PR creates a generic OnboardingScaffold inside Core2 and uses it inside the app


app/src/main/java/com/infomaniak/swisstransfer/ui/OnboardingActivity.kt
```diff
@@ -17,13 +17,51 @@
...
+@AndroidEntryPoint
 class OnboardingActivity : ComponentActivity() {
+
+    @Inject
+    lateinit var accountUtils: AccountUtils
+
     override fun onCreate(savedInstanceState: Bundle?) {
-        enableEdgeToEdge()
         super.onCreate(savedInstanceState)
+        enableEdgeToEdge(navigationBarStyle = SystemBarStyle.auto(Color.TRANSPARENT, Color.TRANSPARENT))
```
KevinBoulongne, line 45: Shouldn't this line be before the `super.onCreate()` ?

kDoc :
image

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/OnboardingScreen.kt
```diff
@@ -0,0 +1,195 @@
...
+    text = {
+        TitleAndDescription(
+            page = this,
+            isHighlighted = { isHighlighted[this]?.value ?: false }
+        )
+    }
+)
+
+@Composable
+private fun Illustration(illustration: ThemedImage) {
```
FabianDevel, line 106: This is only used once and is a oneliner
Is is relly interesting to get it in a separate component here ? 
It feels like a unnecessary wrap of the original Image component

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/OnboardingScreen.kt
```diff
@@ -0,0 +1,195 @@
...
+@Composable
+private fun TitleAndDescription(page: Page, isHighlighted: () -> Boolean) {
+    Column(horizontalAlignment = Alignment.CenterHorizontally) {
+        Text(
+            textAlign = TextAlign.Center,
+            text = stringResource(page.titleRes),
+            style = SwissTransferTheme.typography.specificLight22,
+        )
+
+        Spacer(modifier = Modifier.height(Margin.Mini))
```
FabianDevel, line 119: I see on the maquettes that this title is supposed to be in the same paragraph as the description. 
This mini padding feels a bit strange, because it's not a real title paddng, but not in the same paragraph as the description either... 
And the text is really big, so a mini Margin feels a bit out of place in my opinion

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/OnboardingScreen.kt
```diff
@@ -0,0 +1,195 @@
...
+        )
+    }
+}
+
+@Composable
+private fun BottomContent(
+    modifier: Modifier = Modifier,
+    isLastPage: () -> Boolean,
+    startMainActivity: () -> Unit,
+    goToNextPage: () -> Unit
```
FabianDevel, line 137:
```suggestion
    goToNextPage: () -> Unit,
```

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/OnboardingScreen.kt
```diff
@@ -0,0 +1,195 @@
...
+) {
+    Box(
+        contentAlignment = Alignment.Center,
+        modifier = modifier
+            .fillMaxWidth()
+            .height(140.dp),
```
FabianDevel, line 143: Maybe put this magic value as a private const

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/OnboardingScreen.kt
```diff
@@ -0,0 +1,195 @@
...
+    @StringRes val subtitleArgumentRes: Int,
+    val highlightAngleDegree: Double,
+) {
+    STORAGE(
+        background = AppIllus.RadialGradientCornerTopRight,
+        illustration = AppIllus.StorageBoxPile,
+        titleRes = R.string.onboardingStorageTitle,
+        subtitleTemplateRes = R.string.onboardingStorageSubtitleTemplate,
+        subtitleArgumentRes = R.string.onboardingStorageSubtitleArgument,
+        highlightAngleDegree = -3.0,
```
FabianDevel, line 166: We could extract this to a private const too, especially as all the angle of these highlighted texts should probably always stay the same for this case

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/components/AnimatedOnboardingButton.kt
```diff
@@ -0,0 +1,110 @@
...
+            .padding(horizontal = Margin.Medium)
+            .width(buttonWidth),
+        onClick = onClick,
+        contentPadding = PaddingValues(),
+    ) {
+        Box(contentAlignment = Alignment.Center) {
+            Icon(
+                modifier = Modifier.alpha(arrowVisibility),
+                imageVector = AppImages.AppIcons.ArrowRight,
+                contentDescription = null
```
FabianDevel, line 86: Doesn't this should be something like "next" ? 
Also it misses a comma

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/components/AnimatedOnboardingButton.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    val buttonHeight by animateDpAsState(
+        targetValue = if (isExpanded()) Dimens.LargeButtonHeight else FAB_SIZE,
+        animationSpec = tween(durationMillis = BUTTON_ANIM_DELAY),
+        label = "Onboarding button height",
+    )
+
+    val arrowDelay = if (isExpanded()) BUTTON_ANIM_NO_DELAY else BUTTON_ANIM_DELAY
+    val arrowVisibility by animateFloatAsState(
+        targetValue = if (isExpanded()) 0f else 1f,
```
FabianDevel, line 62: we could use the Boolean.toFloat() here and for textVisibility

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/components/AnimatedOnboardingButton.kt
```diff
@@ -0,0 +1,110 @@
...
+        Box(contentAlignment = Alignment.Center) {
+            Icon(
+                modifier = Modifier.alpha(arrowVisibility),
+                imageVector = AppImages.AppIcons.ArrowRight,
+                contentDescription = null
+            )
+            Text(
+                modifier = Modifier.alpha(textVisibility),
+                text = stringResource(id = R.string.buttonStart),
+                style = SwissTransferTheme.typography.bodyMedium
```
FabianDevel, line 91:
```suggestion
                style = SwissTransferTheme.typography.bodyMedium,
```

Core2/Onboarding/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
...
+#   public *;
+#}
+
+# Uncomment this to preserve the line number information for
+# debugging stack traces.
+#-keepattributes SourceFile,LineNumberTable
+
+# If you keep the line number information, uncomment this to
+# hide the original source file name.
+#-renamesourcefileattribute SourceFile
```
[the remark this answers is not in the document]
KevinBoulongne: Default file generated by Android Studio, we usually don't touch them.
There's already 6 versions of this file, like this.

Core2/Onboarding/build.gradle.kts
```diff
@@ -0,0 +1,51 @@
...
+        kotlinCompilerExtensionVersion = "1.5.14"
+    }
+    kotlinOptions {
+        jvmTarget = sharedJavaVersion.toString()
+    }
+}
+
+dependencies {
+
+    implementation(core2.androidx.core.ktx)
```
KevinBoulongne, line 45:
```suggestion
    implementation(libs.androidx.core.ktx)
```

Core2/Onboarding/src/main/java/com/infomaniak/library/onboarding/OnboardingScaffold.kt
```diff
@@ -0,0 +1,153 @@
...
+    }
+}
+
+data class OnboardingPage(
+    val background: ImageVector,
+    val illustration: @Composable () -> Unit,
+    val text: @Composable () -> Unit,
+)
+
+
```
KevinBoulongne, line 113:
```suggestion
```

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/HighlightedText.kt
```diff
@@ -52,12 +53,62 @@ private const val ROTATION_ANGLE_DEGREE = -3.0
...
     verticalPadding: Float = VERTICAL_PADDING,
     horizontalPadding: Float = HORIZONTAL_PADDING,
     angleDegrees: Double = ROTATION_ANGLE_DEGREE,
+    isHighlighted: () -> Boolean,
+) {
+    HighlightedTextCore(
+        templateRes,
+        argumentRes,
+        style,
+        modifier,
```
KevinBoulongne, line 69:
```suggestion
        modifier,
        templateRes,
        argumentRes,
        style,
```

app/src/main/java/com/infomaniak/swisstransfer/ui/components/HighlightedText.kt
```diff
@@ -52,12 +53,62 @@ private const val ROTATION_ANGLE_DEGREE = -3.0
...
+    angleDegrees: Double = ROTATION_ANGLE_DEGREE,
+) {
+    var isHighlighted by rememberSaveable { mutableStateOf(false) }
+    LaunchedEffect(Unit) { isHighlighted = true }
+
+    HighlightedTextCore(
+        templateRes,
+        argumentRes,
+        style,
+        modifier,
```
KevinBoulongne, line 94:
```suggestion
        modifier,
        templateRes,
        argumentRes,
        style,
```

app/src/main/java/com/infomaniak/swisstransfer/ui/components/HighlightedText.kt
```diff
@@ -52,12 +53,62 @@ private const val ROTATION_ANGLE_DEGREE = -3.0
...
+        { isHighlighted },
+    )
+}
+
+@Composable
+private fun HighlightedTextCore(
+    templateRes: Int,
+    argumentRes: Int,
+    style: TextStyle,
+    modifier: Modifier,
```
KevinBoulongne, line 107:
```suggestion
    modifier: Modifier,
    templateRes: Int,
    argumentRes: Int,
    style: TextStyle,
```

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/OnboardingActivity.kt
LunarX: What matters is only calling it before `setContent`. `super.onCreate` doesn't actually matter. We can indeed call it before, but like most other lines of code. You can also observe that when creating a new empty project in Android Studio the line is also placed after `super.onCreate`. 

If not important, I'd rather have it called after `super.onCreate` because, when you see a line of code before `super.onCreate` it makes you believe that placing it here was absolutely required, which is not the case here.

Capture d’écran 2024-12-03 à 13 07 11
https://developer.android.com/develop/ui/views/layout/edge-to-edge#enable-edge-to-edge-display

app/src/main/java/com/infomaniak/swisstransfer/ui/images/illus/onboarding/StorageBoxPileDark.kt
```diff
@@ -0,0 +1,632 @@
...
+            viewportWidth = 289.0f,
+            viewportHeight = 258.0f
+        ).apply {
+            path(
+                fill = SolidColor(Color(0xFFA98116)),
+                stroke = null,
+                strokeLineWidth = 0.0f,
+                strokeLineCap = Butt,
+                strokeLineJoin = Miter,
+                strokeLineMiter = 4.0f,
```
KevinBoulongne, line 43: Stroke parameters can be removed for a lot of paths in this file.

app/src/main/java/com/infomaniak/swisstransfer/ui/images/illus/onboarding/StorageBoxPileDark.kt
```diff
@@ -0,0 +1,632 @@
...
+            ) {
+                moveTo(97.98f, 94.05f)
+                curveToRelative(0.78f, 0.68f, 3.34f, 1.72f, 4.57f, 1.98f)
+                curveToRelative(-1.29f, -0.57f, -3.01f, -1.2f, -3.95f, -1.99f)
+                curveToRelative(-4.49f, -3.76f, -1.84f, -10.04f, 3.95f, -9.12f)
+                curveToRelative(1.06f, 0.17f, 2.86f, 1.02f, 4.31f, 1.73f)
+                curveToRelative(-0.97f, -0.69f, -3.81f, -2.05f, -4.84f, -2.28f)
+                curveToRelative(-4.83f, -1.07f, -9.7f, 4.74f, -4.04f, 9.68f)
+            }
+            group {
```
KevinBoulongne, line 54: This group looks like it can be removed.

app/src/main/java/com/infomaniak/swisstransfer/ui/images/illus/onboarding/StorageBoxPileDark.kt
```diff
@@ -0,0 +1,632 @@
...
+                    strokeLineJoin = Miter,
+                    strokeLineMiter = 4.0f,
+                    pathFillType = EvenOdd
+                ) {
+                    moveTo(113.15f, 103.25f)
+                    curveToRelative(-0.14f, -0.05f, -10.5f, 3.16f, -7.58f, 7.85f)
+                    curveToRelative(-0.47f, -2.07f, 0.5f, -4.47f, 7.58f, -7.85f)
+                }
+            }
+            group {
```
KevinBoulongne, line 144: This group looks like it can be removed.

app/src/main/java/com/infomaniak/swisstransfer/ui/images/illus/onboarding/StorageBoxPileLight.kt
```diff
@@ -0,0 +1,633 @@
...
+        _storageBoxPileLight = Builder(
+            name = "StorageBoxPileLight",
+            defaultWidth = 290.0.dp,
+            defaultHeight = 258.0.dp,
+            viewportWidth = 290.0f,
+            viewportHeight = 258.0f
+        ).apply {
+            path(
+                fill = SolidColor(Color(0xFFA98116)),
+                stroke = null,
```
KevinBoulongne, line 39: Stroke parameters can be removed for a lot of paths in this file.

app/src/main/java/com/infomaniak/swisstransfer/ui/images/illus/onboarding/StorageBoxPileLight.kt
```diff
@@ -0,0 +1,633 @@
...
+            ) {
+                moveTo(98.03f, 94.1f)
+                curveToRelative(0.78f, 0.68f, 3.34f, 1.72f, 4.57f, 1.98f)
+                curveToRelative(-1.3f, -0.57f, -3.01f, -1.2f, -3.95f, -1.99f)
+                curveToRelative(-4.49f, -3.77f, -1.84f, -10.05f, 3.96f, -9.13f)
+                curveToRelative(1.06f, 0.17f, 2.86f, 1.02f, 4.31f, 1.73f)
+                curveToRelative(-0.97f, -0.69f, -3.82f, -2.05f, -4.84f, -2.28f)
+                curveToRelative(-4.83f, -1.07f, -9.71f, 4.74f, -4.05f, 9.68f)
+            }
+            group {
```
KevinBoulongne, line 54: This group looks like it's useless, and it's not the only one in this file.

app/src/main/java/com/infomaniak/swisstransfer/ui/images/illus/onboarding/StorageBoxPileLight.kt
KevinBoulongne: It's probably the same in the following illustrations, I'll let you check them.

app/src/main/java/com/infomaniak/swisstransfer/ui/images/illus/onboarding/StorageBoxPileLight.kt
KevinBoulongne: It's probably the same in the following illustrations, I'll let you check them.

KevinBoulongne (changes requested):

Core2/Onboarding/src/main/java/com/infomaniak/library/onboarding/OnboardingScaffold.kt
```diff
@@ -0,0 +1,153 @@
...
+                    start = paddingValues.calculateStartPadding(LocalLayoutDirection.current),
+                    end = paddingValues.calculateEndPadding(LocalLayoutDirection.current),
+                )
+            )
+        }
+    }
+}
+
+@Composable
+private fun OnboardingPageContent(modifier: Modifier, page: OnboardingPage, calculateTopPadding: Dp) {
```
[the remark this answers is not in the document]
LunarX: Yeah, this actually hid some more problems. In the end, I only needed the top padding and it's not passed as a modifier anymore

Core2/Onboarding/src/main/java/com/infomaniak/library/onboarding/OnboardingScaffold.kt
```diff
@@ -0,0 +1,153 @@
...
+
+@Composable
+fun OnboardingScaffold(
+    pagerState: PagerState,
+    onboardingPages: List<OnboardingPage>,
+    bottomContent: @Composable (PaddingValues) -> Unit,
+) {
+    Scaffold { paddingValues ->
+        Column {
+            val left = paddingValues.calculateLeftPadding(LocalLayoutDirection.current)
```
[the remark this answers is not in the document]
LunarX: The start/end and left/right thing did not seem like a very good approach, I managed to only use PaddingValues instances instead to solve this inconsistency.

The insets can't be set to the `Column` nor the `HorizontalPager` because I need the background part of the `OnboardingPageContent` to draw behind the system bars (at the top of the screen and horizontally). I also need the `HorizontalPager`'s content to draw inside the "safe areas" by default but have them drawn behind the system bars when they're scrolled (vertically or horizontally)

Core2/Onboarding/src/main/java/com/infomaniak/library/onboarding/OnboardingScaffold.kt
```diff
@@ -0,0 +1,153 @@
...
+            val right = paddingValues.calculateRightPadding(LocalLayoutDirection.current)
+
+            HorizontalPager(pagerState, modifier = Modifier.weight(1f)) {
+                OnboardingPageContent(
+                    modifier = Modifier.windowInsetsPadding(
+                        WindowInsets(left = left, top = paddingValues.calculateTopPadding(), right = right)
+                    ),
+                    page = onboardingPages[it],
```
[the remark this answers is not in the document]
LunarX: Yeah it definitely doesn't hurt

Core2/Onboarding/src/main/java/com/infomaniak/library/onboarding/OnboardingScaffold.kt
```diff
@@ -0,0 +1,153 @@
...
+@Composable
+private fun Preview() {
+    val imageVector = Builder(
+        defaultWidth = 24.0.dp,
+        defaultHeight = 24.0.dp,
+        viewportWidth = 24.0f,
+        viewportHeight = 24.0f,
+    ).build()
+
+    val a = OnboardingPage(
```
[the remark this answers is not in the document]
LunarX: Yeah I forgot I used a temporary name for my variable, good catch

app/src/main/java/com/infomaniak/swisstransfer/ui/OnboardingActivity.kt
```diff
@@ -17,13 +17,51 @@
...
+        setContent {
+            val scope = rememberCoroutineScope()
+
+            SwissTransferTheme {
+                Surface {
+                    OnboardingScreen(
+                        goToMainActivity = {
+                            scope.launch {
+                                accountUtils.login()
+                                startActivity(Intent(this@OnboardingActivity, MainActivity::class.java))
```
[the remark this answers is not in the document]
LunarX: Could we rather simplify the writing of trivial `startActivity`s with something like this instead if we want an easier way of reading code?:

```kt
startActivity(MainActivity::class)
```

This can be achieved using the following method:

```kt
fun <T : Activity> Context.startActivity(targetActivity: KClass<T>) {
    startActivity(Intent(this, targetActivity.java))
}
```

Core2/Onboarding/src/main/java/com/infomaniak/library/onboarding/OnboardingScaffold.kt
```diff
@@ -0,0 +1,153 @@
...
+                OnboardingPageContent(
+                    modifier = Modifier.windowInsetsPadding(
+                        WindowInsets(left = left, top = paddingValues.calculateTopPadding(), right = right)
+                    ),
+                    page = onboardingPages[it],
+                    0.dp,
+                )
+            }
+
+            HorizontalPagerIndicator(
```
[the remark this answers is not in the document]
LunarX: I figured, when you're in landscape mode you're already struggling a lot with vertical space which is why I was thinking, I'd rather not put vertical margins to not loose more space. Instead of having something nice at least you can see a bit more info on the screen

app/src/main/java/com/infomaniak/swisstransfer/ui/components/HighlightedText.kt
```diff
@@ -52,12 +53,62 @@ private const val ROTATION_ANGLE_DEGREE = -3.0
...
+        modifier,
+        verticalPadding,
+        horizontalPadding,
+        angleDegrees,
+        isHighlighted,
+    )
+}
+
+@Composable
+fun HighlightedText(
```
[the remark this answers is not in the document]
LunarX: Yeah I was wondering the same thing, but yeah I agree, I will remove the 2nd signature, it' will be more intuitive

app/src/main/java/com/infomaniak/swisstransfer/ui/components/HighlightedText.kt
```diff
@@ -66,12 +117,9 @@ fun HighlightedText(
     val highlightedColor = SwissTransferTheme.colors.highlightedColor
 
     var boundingBoxes by remember { mutableStateOf<List<Rect>>(emptyList()) }
-    var animationStarted by rememberSaveable { mutableStateOf(false) }
-
-    LaunchedEffect(Unit) { animationStarted = true }
 
     val highlightProgress by animateFloatAsState(
-        targetValue = if (animationStarted) 1f else 0f,
+        targetValue = if (isHighlighted()) 1f else 0f,
```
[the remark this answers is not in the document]
LunarX: Here I'd find it slightly less clear if we used `.toFloat`. Currently I can really express that when it is highlighted I want to have progress be filled at 100% and when it's not have it be filled at 0%. And it so happens to be the exact same values as `.toFloat`. But using `.toFloat` would change the meaning of the line and not carry this "fully progressed" vs "not progressed at all", don't you think?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/OnboardingScreen.kt
```diff
@@ -0,0 +1,195 @@
...
+    val isLastPage by remember { derivedStateOf { pagerState.currentPage >= pagerState.pageCount - 1 } }
+    val coroutineScope = rememberCoroutineScope()
+
+    val isHighlighted = Page.entries.associateWith { rememberSaveable { mutableStateOf(false) } }
+
+    // Start the highlighting of the text when the associated page is reached in the HorizontalPager
+    LaunchedEffect(pagerState.currentPage) {
+        val currentPage = Page.entries[pagerState.currentPage]
+        val setIsHighlighted = isHighlighted[currentPage]?.component2()
+        setIsHighlighted?.invoke(true)
```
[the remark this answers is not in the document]
LunarX: Yes, I thought this part was way too many back and forth but I couldn't see how to do it easily yet it was so simple

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/OnboardingScreen.kt
```diff
@@ -0,0 +1,195 @@
...
+        pagerState = pagerState,
+        onboardingPages = onboardingPages,
+        bottomContent = {
+            BottomContent(
+                modifier = Modifier
+                    .padding(it)
+                    .consumeWindowInsets(it),
+                isLastPage = { isLastPage },
+                startMainActivity = goToMainActivity,
+                goToNextPage = { coroutineScope.launch { pagerState.animateScrollToPage(pagerState.currentPage + 1) } },
```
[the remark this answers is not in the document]
LunarX: My point of view was that the `OnboardingScreen` calls the `OnboardingScaffold` and dispatches the responsibilities to each component and that's pretty much it. 

Then the choice of the click was handled inside its dedicated component. 

Also does it not help with limiting recompositions?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/OnboardingScreen.kt
LunarX: It does indeed, but without it the green part of the highlighted text hides the title because the title is drawn before the description. Apparently there's no way to change the drawing order inside of a Column, it's always top to bottom :/

But I'm not sure what else we can do apart from spacing it a bit

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/OnboardingScreen.kt
LunarX: I don't really know if we need to name all dimensions that are only used once

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/components/AnimatedOnboardingButton.kt
LunarX: Very true indeed, I'll add a content description

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/components/AnimatedOnboardingButton.kt
LunarX: Same as the other .toFloat(), I feel like we loose part of the meaning, even more here

Core2/gradle/core2.versions.toml
```diff
@@ -1,7 +1,19 @@
 [versions]
-sentry-android = "7.15.0"
+coreKtx = "1.15.0"
+foundationAndroid = "1.7.3"
+runtimeAndroid = "1.7.5"
+material3Android = "1.3.1"
+uiTooling = "1.7.5"
+uiToolingPreviewAndroid = "1.7.5"
```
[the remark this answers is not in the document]
LunarX: Yes correct. I checked and introduced the same compose BOM we use in the swiss transfer project to manage the compose dependencies versions

app/src/main/java/com/infomaniak/swisstransfer/ui/images/illus/onboarding/TwoPadlocksIntertwinedStars.kt
```diff
@@ -0,0 +1,63 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.layout.ContentScale
+import androidx.compose.ui.tooling.preview.Preview
+import com.infomaniak.swisstransfer.ui.images.AppImages
+import com.infomaniak.swisstransfer.ui.images.AppImages.AppIllus
+import com.infomaniak.swisstransfer.ui.images.ThemedImage
+import com.infomaniak.swisstransfer.ui.theme.SwissTransferTheme
+
+val AppIllus.TwoPadlocksIntertwinedStars: ThemedImage
```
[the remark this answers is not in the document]
LunarX: Yeah both names are synonyms imo, whichever should be fine, I'll rename it

app/src/main/java/com/infomaniak/swisstransfer/ui/images/illus/onboarding/RadialGradientCornerTopRightLight.kt
```diff
@@ -0,0 +1,71 @@
...
+            return _radialGradientCornerTopRightLight!!
+        }
+        _radialGradientCornerTopRightLight = Builder(
+            name = "RadialGradientCornerTopRightLight",
+            defaultWidth = 375.0.dp,
+            defaultHeight = 832.0.dp,
+            viewportWidth = 375.0f,
+            viewportHeight = 832.0f
+        ).apply {
+            group {
```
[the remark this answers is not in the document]
LunarX: There's *a lot* of groups that can be removed, I don't think we want to loose too much time on this. It takes long enough to reimport all illustrations each time something changes to not bother with going through multiple 500~2000 lines illustrations to check if every group is actually needed

app/src/main/java/com/infomaniak/swisstransfer/ui/images/illus/onboarding/StorageBoxPileLight.kt
LunarX: Yeah I removed all null stroke I found in the illustrations

app/src/main/java/com/infomaniak/swisstransfer/ui/images/illus/onboarding/StorageBoxPileDark.kt
LunarX: About groups in generaé, I have answered to the same comment here: https://github.com/Infomaniak/android-SwissTransfer/pull/222#discussion_r1870905083

app/src/main/java/com/infomaniak/swisstransfer/ui/images/illus/onboarding/StorageBoxPileDark.kt
LunarX: About groups in generaé, I have answered to the same comment here: https://github.com/Infomaniak/android-SwissTransfer/pull/222#discussion_r1870905083

app/src/main/java/com/infomaniak/swisstransfer/ui/images/illus/onboarding/StorageBoxPileLight.kt
LunarX: About groups in general, I have answered to the same comment here: https://github.com/Infomaniak/android-SwissTransfer/pull/222#discussion_r1870905083

Core2/gradle/core2.versions.toml
```diff
@@ -1,7 +1,16 @@
...
 
 [libraries]
-sentry-android = { module = "io.sentry:sentry-android", version.ref = "sentry-android" }
+androidx-core-ktx = { group = "androidx.core", name = "core-ktx", version.ref = "coreKtx" }
+compose-foundation = { group = "androidx.compose.foundation", name = "foundation-android" }
+compose-material3 = { group = "androidx.compose.material3", name = "material3-android" }
+compose-runtime = { group = "androidx.compose.runtime", name = "runtime-android" }
+compose-ui-tooling = { group = "androidx.compose.ui", name = "ui-tooling" }
+compose-ui-tooling-preview = { group = "androidx.compose.ui", name = "ui-tooling-preview-android" }
+compose-bom = { group = "androidx.compose", name = "compose-bom", version.ref = "composeBom" }
```
KevinBoulongne, line 14: Almost all of them are already available in `libs`, no need to add them in `core2`.

Core2/Onboarding/build.gradle.kts
KevinBoulongne: You resolved the conversation but didn't update the code.

app/src/main/java/com/infomaniak/swisstransfer/ui/MainApplication.kt
```diff
@@ -59,7 +59,7 @@ class MainApplication : Application(), Configuration.Provider {
 
         globalCoroutineScope.launch {
             accountUtils.init()
-            transferManager.deleteExpiredTransfers()
+            if (accountUtils.isUserConnected()) transferManager.deleteExpiredTransfers()
```
KevinBoulongne, line 62: Why this change ?

KevinBoulongne (changes requested):

Core2/Onboarding/build.gradle.kts
LunarX: Not sure how I understood your comment the first time because I added a "thumbs up" but actually we don't want to depend on the SwissTransfer module inside of Core2

Core2/gradle/core2.versions.toml
LunarX: We want them in Core2 to have the Core2 module be independent from the SwissTransfer module.

An alternative though could be to define SwissTransfer's dependencies based on the Core2 ones. Maybe this needs to be thought through just to be sure but this would recreate the same ecosystem we had with the original Core as well

app/src/main/java/com/infomaniak/swisstransfer/ui/MainApplication.kt
LunarX: `deleteExpiredTransfers` crashes when you start the app because it can't find any realm to read from when there's no user connected. And at this point in time, during your very first start of the app, you're not connected to any account yet

Core2/src/main/java/com/infomaniak/core2/ScreenUtils.kt
```diff
@@ -0,0 +1,40 @@
...
+        (heightPixels / ydpi) to (widthPixels / xdpi)
+    }
+
+    val aspectRatio = resources.configuration.screenLayout and Configuration.SCREENLAYOUT_LONG_MASK
+    val isLongScreen = aspectRatio == Configuration.SCREENLAYOUT_LONG_YES
+
+    val isScreenTooSmall = isLongScreen && min(screenHeightInches, screenWidthInches) < MIN_HEIGHT_FOR_LANDSCAPE
+
+    if (isScreenTooSmall) requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_PORTRAIT
+}
```
KevinBoulongne, line 40: We already have some code to know if the screen is small or not.
We should reuse it.
It's in `WindowAdaptativeInfoExt`.

KevinBoulongne (changes requested):

Core2/src/main/java/com/infomaniak/core2/ScreenUtils.kt
KevinBoulongne: https://stackoverflow.com/questions/69230049/how-to-force-orientation-for-some-screens-in-jetpack-compose

LunarX: I based the logic on the compose method we already use everywhere else to be sure we are consistent throughout the app 👍

Core2/gradle/core2.versions.toml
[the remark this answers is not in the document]
LunarX: I tried doing so but apparently I stopped before finishing. Anyway, I reordered them correctly

Core2/src/main/java/com/infomaniak/core2/ScreenUtils.kt
```diff
@@ -0,0 +1,40 @@
...
+ */
+package com.infomaniak.core2
+
+import android.annotation.SuppressLint
+import android.app.Activity
+import android.content.pm.ActivityInfo
+import android.content.res.Configuration
+import kotlin.math.min
+
+private const val MIN_HEIGHT_FOR_LANDSCAPE = 4
```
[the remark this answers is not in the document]
LunarX: I agree, but in the end this code was replaced and the constant is not needed anymore

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/OnboardingScreen.kt
```diff
@@ -0,0 +1,188 @@
...
+    OnboardingScaffold(
+        pagerState = pagerState,
+        onboardingPages = onboardingPages,
+        bottomContent = { paddingValues ->
+            BottomContent(
+                modifier = Modifier
+                    .padding(paddingValues)
+                    .consumeWindowInsets(paddingValues),
+                isLastPage = { isLastPage },
+                startMainActivity = goToMainActivity,
```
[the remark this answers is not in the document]
LunarX: Oh yes! I wanted to rename that one as `goToMainActivity` everywhere but I went about my day and forgot, thanks I fixed it

Core2/Onboarding/src/main/java/com/infomaniak/library/onboarding/OnboardingScaffold.kt
```diff
@@ -0,0 +1,152 @@
...
+}
+
+data class OnboardingPage(
+    val background: ImageVector,
+    val illustration: @Composable () -> Unit,
+    val text: @Composable () -> Unit,
+)
+
+@Preview
+@Preview(device = "spec:parent=pixel_5,orientation=landscape")
```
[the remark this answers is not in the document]
LunarX: The previews from SwissTransfer are not accessible from this module unfortunately :/

tevincent (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/OnboardingScreen.kt
```diff
@@ -0,0 +1,188 @@
...
+import com.infomaniak.swisstransfer.ui.theme.Margin
+import com.infomaniak.swisstransfer.ui.theme.SwissTransferTheme
+import com.infomaniak.swisstransfer.ui.utils.PreviewLargeWindow
+import com.infomaniak.swisstransfer.ui.utils.PreviewSmallWindow
+import kotlinx.coroutines.launch
+
+@Composable
+fun OnboardingScreen(goToMainActivity: () -> Unit) {
+    val pagerState = rememberPagerState(pageCount = { Page.entries.size })
+    val isLastPage by remember { derivedStateOf { pagerState.currentPage >= pagerState.pageCount - 1 } }
```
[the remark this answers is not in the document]
LunarX: It did look great but after further testing, it doesn't work like we want it :/

tevincent (approved):

Core2/Onboarding/src/main/java/com/infomaniak/library/onboarding/OnboardingScaffold.kt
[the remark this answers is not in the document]
LunarX: Ultimately we definitely should, but instead of moving half the project to Core2 I tried limiting the modifications to what felt like the most important parts. 

Do we deem the previews important enough and I move them already here?

app/build.gradle.kts
```diff
@@ -88,6 +88,7 @@ dependencies {
     implementation(project(":Core2:Matomo"))
     implementation(project(":Core2:Network"))
     implementation(project(":Core2:Onboarding"))
+    implementation(project(":Core2:compose:Core"))
```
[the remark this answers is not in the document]
LunarX: Yes, I agree, I renamed it

FabianDevel (approved):

Core2/Onboarding/src/main/java/com/infomaniak/library/onboarding/OnboardingScaffold.kt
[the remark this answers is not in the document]
FabianDevel: Finally we decided to lock the screen in portrait mode for small device to avoid these problems

Core2/Onboarding/src/main/java/com/infomaniak/library/onboarding/OnboardingScaffold.kt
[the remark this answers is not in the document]
KevinBoulongne: The Preview's code is small enough and generic enough to be moved now, I guess.
It could be in another PR though.

Core2/src/main/java/com/infomaniak/core2/ScreenUtils.kt
KevinBoulongne: Perfect.
This file can be deleted now.

Core2/Onboarding/src/main/java/com/infomaniak/library/onboarding/OnboardingScaffold.kt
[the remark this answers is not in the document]
LunarX: We create a card to do this task later 👍

sirambd (changes requested):

sirambd (approved):

PR #227, opened 2024-12-03
feat: Add correct splashscreen for all versions

KevinBoulongne:
- Before 31
- 31+ in Light
- 31+ in Dark

image


app/src/main/res/drawable/splashscreen_legacy.xml
```diff
@@ -0,0 +1,17 @@
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
```
KevinBoulongne, line 1: The header is missing, it's because of a weird Git diff.

It's added here : #228

tevincent (changes requested):
Maybe we can use `webp`s ? Just tried converting all `png`s to `webp`s and it seems to working fine. That way, we're saving ~50% in size.

PR #230, opened 2024-12-04
feat: Add `String.isEmail()` function

KevinBoulongne:


Core2/src/main/java/com/infomaniak/core2/EmailUtils.kt
```diff
@@ -0,0 +1,22 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.infomaniak.core2
+
+import android.util.Patterns
+
+fun String.isEmail(): Boolean = Patterns.EMAIL_ADDRESS.matcher(this).matches()
```
tevincent, line 22: Maybe we can add this to KMP ? That way, we can use the same regex for both platform ?

tevincent (changes requested):

Core2/src/main/java/com/infomaniak/core2/EmailUtils.kt
KevinBoulongne: `Patterns` is in the Android SDK, it's not available in KMP.

tevincent (approved):

PR #232, opened 2024-12-04
feat: Make sure Transfer author email is mandatory and correctly formatted

KevinBoulongne:
Depends on #220


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/ImportFilesViewModel.kt
```diff
@@ -240,6 +240,11 @@ class ImportFilesViewModel @Inject constructor(
     }
     //endregion
 
+    //region Transfer Author Email
+    private var _transferAuthorEmail by mutableStateOf("")
+    val transferAuthorEmail = GetSetCallbacks(get = { _transferAuthorEmail }, set = { _transferAuthorEmail = it })
+    //endregion
+
```
tevincent, line 247: Maybe we can take this opportunity to move all variable here to the top with the rest ? I think they want to be together again.

tevincent (changes requested):

tevincent (approved):

PR #237, opened 2024-12-05
feat: Create and style correctly the HorizontalPagerIndicator for the onboarding

LunarX:
Depends on #222


Core2/Onboarding/src/main/java/com/infomaniak/library/onboarding/OnboardingScaffold.kt
```diff
@@ -147,6 +152,13 @@ private fun Preview() {
...
                 Text("Bottom content")
             }
-        }
+        },
+        indicatorStyle = IndicatorStyle(
+            inactiveColor = Color.LightGray,
+            activeColor = Color.DarkGray,
+            inactiveSize = 8.dp,
+            activeWidth = 16.dp,
+            indicatorSpacing = 8.dp,
```
tevincent, line 161: Any reason to not use `Margin.Mini` here ?

Core2/Onboarding/src/main/java/com/infomaniak/library/onboarding/OnboardingScaffold.kt
```diff
@@ -61,8 +62,12 @@ fun OnboardingScaffold(
             }
 
             HorizontalPagerIndicator(
-                modifier = Modifier.padding(PaddingValues(start = startPadding, end = endPadding)),
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .padding(vertical = 8.dp)
```
tevincent, line 67: You can also use `Margin.Mini` here maybe ?

tevincent (changes requested):

Core2/Onboarding/src/main/java/com/infomaniak/library/onboarding/OnboardingScaffold.kt
LunarX: Yeah, they're not accessible because this is its own submodule that needs to be independent from SwissTransfer where the margins are defined. Ideally the margins definition should be moved to core altogether at some point probably

KevinBoulongne:
The dots color are wrong, if we refer to the design, only the selected page dot should be green.
image
image

PR #253, opened 2024-12-13
feat: Update logos

LunarX:
This PR adds the correct launcher icon and notification icon. The splashscreen icon is still missing and we're waiting for it


tevincent (approved):

tevincent (changes requested):
We should have a `k` in the icon right ?

Capture d’écran 2024-12-16 à 11 01 53

tevincent (approved):

PR #254, opened 2024-12-16
feat: Mail validation screen

LunarX:
Adds the screen to validate the email.

Is still missing: 
* A more robust "open mail app" button

Depends on: https://github.com/Infomaniak/multiplatform-SwissTransfer/pull/133


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/validateemail/ValidateUserEmailScreen.kt
```diff
@@ -0,0 +1,239 @@
...
+
+@Composable
+fun ValidateUserEmailScreen(
+    closeActivity: () -> Unit,
+    navigateBack: () -> Unit,
+    emailToValidate: String,
+    validateUserEmailViewModel: ValidateUserEmailViewModel = hiltViewModel<ValidateUserEmailViewModel>(),
+) {
+    val isLoading by validateUserEmailViewModel.isLoading.collectAsStateWithLifecycle()
+    val isError by validateUserEmailViewModel.isError.collectAsStateWithLifecycle()
```
tevincent, line 69: Maybe we can use the `UiState` pattern here ?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/validateemail/ValidateUserEmailViewModel.kt
```diff
@@ -0,0 +1,79 @@
...
+import dagger.hilt.android.qualifiers.ApplicationContext
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+
+@HiltViewModel
+class ValidateUserEmailViewModel @Inject constructor(
+    @ApplicationContext private val appContext: Context,
```
tevincent, line 37: This is not used.

tevincent (changes requested):

tevincent (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/validateemail/ValidateUserEmailViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    private val _uiState = MutableStateFlow(ValidateEmailUiState.Default)
+    val uiState = _uiState.asStateFlow()
+
+    fun validateEmailWithOtpCode(email: String, otpCode: String, onSuccess: () -> Unit, onUnknownError: () -> Unit) {
+        viewModelScope.launch(ioDispatcher) {
+            _uiState.value = ValidateEmailUiState.Loading
+
+            runCatching {
+                uploadRepository.verifyEmailCode(VerifyEmailCodeBody(otpCode, email))
```
tevincent, line 50: Since you're using the `uploadManager` here, maybe we can add a method in  `uploadManager` to call the `verifyEmailCode` ? `uploadManager` is used as a common access point for the repository and the controller of uploads so it might be cleaner ?

And if we're a doing that on KMP, we can directly save the token in DB, instead of doing that line 57. Let me know if it'll cause any issue with iOS, as we discussed yesterday as it might interfere with what they did on the native iOS side of things.

Also, we're using the manager to do `uploadManager.resendEmailCode(email)` so it might be good to do the same thing for everything else.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/validateemail/components/OtpTextField.kt
```diff
@@ -0,0 +1,232 @@
...
+) {
+    companion object {
+        @Composable
+        fun default(
+            activeColor: Color = MaterialTheme.colorScheme.primary,
+            inactiveColor: Color = MaterialTheme.colorScheme.outline,
+            errorColor: Color = MaterialTheme.colorScheme.error,
+            activeBorderThickness: Dp = 2.dp,
+            inactiveBorderThickness: Dp = 1.dp,
+            shape: RoundedCornerShape = RoundedCornerShape(8.dp),
```
tevincent, line 193:
```suggestion
            shape: RoundedCornerShape = CustomShapes.SMALL,
```

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/validateemail/components/CodeVerification.kt
```diff
@@ -0,0 +1,110 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.alpha
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.TextStyle
+import com.infomaniak.swisstransfer.R
+import com.infomaniak.swisstransfer.ui.theme.Margin
+import com.infomaniak.swisstransfer.ui.theme.SwissTransferTheme
+import com.infomaniak.swisstransfer.ui.utils.PreviewLightAndDark
+
+private val VALID_CHARACTERS = setOf('0', '1', '2', '3', '4', '5', '6', '7', '8', '9')
```
tevincent, line 35: Maybe a CharRange ?

```kotlin
private val VALID_CHARACTERS = '0' .. '9'
```

I tried and it seems to work. Instead of specifying all characters from 0 to 9.

tevincent (changes requested):
Capture d’écran 2024-12-18 à 08 24 09

Is it me or the padding between the two buttons and the bottom button with the bottom of the screen is not the same ? Is that intended ?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/validateemail/components/OtpTextField.kt
LunarX: I tried making this default style independent of swiss transfer so we can easily extract the otp code component and reuse it elsewhere if needed. But I did override the shape where I use it instead, so it correctly depends on CustomShapes.SMALL like you proposed. Is this ok with you?

LunarX:
> Is it me or the padding between the two buttons and the bottom button with the bottom of the screen is not the same ? Is that intended ?

It's indeed not the same padding. But was it intended? No one really thought about this I think when we changed the vertical outer paddings

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/validateemail/ValidateUserEmailViewModel.kt
LunarX: I had the exact same point of view, we should access the manager rather than the repository. I ended up changing it this way because I needed to add some logic as well, so yeah problem solved :)

We first chose to use the uploadRepository just because no further processing had to be done and it was deemed easier to use what was already there than expose two methods through the manager that don't do anything apart from calling the repository

sirambd (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/validateemail/components/OtpTextField.kt
tevincent: Yes I'm ok with that.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/validateemail/ValidateUserEmailScreen.kt
```diff
@@ -0,0 +1,266 @@
...
+        }
+    }
+}
+
+@Composable
+fun HandleUnknownValidationError(uiState: () -> ValidateEmailUiState, snackbarHostState: SnackbarHostState) {
+    val scope = rememberCoroutineScope()
+    val context = LocalContext.current
+
+    LaunchedEffect(uiState()) {
```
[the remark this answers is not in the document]
LunarX: LaunchedEffect seems to still be needed in order to call the snackbar inside a coroutine

sirambd (approved):

PR #255, opened 2024-12-16
feat: Add confirmation when trying to leave the new transfer flow

LunarX:
Don't let them escape!

This required me to refactor the dialog composable we already have to have greater control over the buttons. I opted for a more standard interfacing logic instead of adding a *lot* of parameters. We can now compose the dialog with buttons. There's still two constant buttons already defined for the standard cases.

This PR also required to add stacking buttons in dialogs when the buttons are too wide which required to change the design of the buttons for plain text buttons like the native material theme so they are not misaligned when stacked.

Depends on #254

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
@@ -35,7 +35,7 @@ object SwissTransferAlertDialogDefaults {
     @Composable
     fun ConfirmButton(isEnabled: () -> Boolean = { true }, onClick: () -> Unit) {
         SmallButton(
-            style = ButtonType.TERTIARY,
+            style = ButtonType.Tertiary,
             title = stringResource(R.string.buttonConfirm),
             enabled = isEnabled,
             onClick = onClick,
@@ -45,7 +45,7 @@ object SwissTransferAlertDialogDefaults {
     @Composable
     fun CancelButton(onClick: () -> Unit) {
         SmallButton(
-            style = ButtonType.TERTIARY,
+            style = ButtonType.Tertiary,
             title = stringResource(RCore2.string.buttonCancel),
             onClick = onClick,
         )
@@ -157,7 +157,7 @@ private fun WideButtonsPreview() {
             SwissTransferAlertDialog(
                 titleRes = R.string.settingsOptionPassword,
                 descriptionRes = R.string.settingsPasswordDescription,
-                positiveButton = { SmallButton("A very looong and wide button", onClick = {}, style = ButtonType.TERTIARY) },
+                positiveButton = { SmallButton("A very looong and wide button", onClick = {}, style = ButtonType.Tertiary) },
                 negativeButton = { SwissTransferAlertDialogDefaults.CancelButton { } },
                 onDismiss = {},
             )
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
@@ -116,15 +116,17 @@ private fun TitleAndDescription(titleRes: Int, descriptionRes: Int) {
     )
 }
 
+@OptIn(ExperimentalLayoutApi::class)
 @Composable
 private fun ActionButtons(
     positiveButton: @Composable () -> Unit,
     negativeButton: @Composable () -> Unit,
 ) {
-    Row(
+    FlowRow(
         modifier = Modifier.fillMaxWidth(),
-        horizontalArrangement = Arrangement.spacedBy(Margin.Micro, Alignment.End),
-        verticalAlignment = Alignment.CenterVertically,
+        horizontalArrangement = Arrangement.spacedBy(Margin.Mini, Alignment.End),
+        verticalArrangement = Arrangement.spacedBy(Margin.Mini),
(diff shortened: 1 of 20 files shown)


tevincent (changes requested):
If the screen is too small, the text is not visible on the right.

Capture d’écran 2024-12-16 à 12 43 54

And pressing the back button should also show the dialog right ?

tevincent (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferButton.kt
```diff
@@ -124,6 +124,12 @@ enum class ButtonType(val buttonColors: @Composable () -> ButtonColors) {
             contentColor = SwissTransferTheme.materialColors.onError,
         )
     }),
+    ERROR_TEXT({
```
FabianDevel, line 127: This is not really an error text, more of a destructive action button

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -97,6 +98,8 @@ fun ImportFilesScreen(
 
     LaunchedEffect(Unit) { importFilesViewModel.initTransferOptionsValues() }
 
+    BackHandler { closeActivity() } // Closing the activity like other screens will prompt the user to be sure he wants to leave
```
FabianDevel, line 101: I was a bit confuse with this sentence at first

Maybe we could rewrite this along the line of "We need to overwrite this to avoid prompting the user to be sure he wants to leave" ?

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
```diff
@@ -31,15 +31,36 @@ import com.infomaniak.swisstransfer.ui.theme.Margin
...
+}
+
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun SwissTransferAlertDialog(
     modifier: Modifier = Modifier,
     @StringRes titleRes: Int,
     @StringRes descriptionRes: Int,
+    positiveButton: @Composable () -> Unit,
+    negativeButton: @Composable () -> Unit,
```
FabianDevel, line 62: This should already use by default your newly created default buttons
I'd suggest to make this nullable all the way to ActionButton, and then manage it there

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
LunarX: I would've loved to as well, but we need to define the on click callbacks outside of it, so I can't set them as default values. The only option would be to provide two signatures for the `SwissTransferAlertDialog`; one that takes composables and one that takes only onClick callbacks. But this feels a bit complicated :/

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
```diff
@@ -31,15 +31,36 @@ import com.infomaniak.swisstransfer.ui.theme.Margin
 import com.infomaniak.swisstransfer.ui.theme.SwissTransferTheme
 import com.infomaniak.core2.R as RCore2
 
+object SwissTransferAlertDialogDefaults {
+    @Composable
+    fun confirmButton(isEnabled: () -> Boolean = { true }, onClick: () -> Unit) {
+        SmallButton(
+            style = ButtonType.TERTIARY,
+            title = stringResource(R.string.buttonConfirm),
```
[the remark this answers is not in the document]
LunarX: We definitely need to, but for now iOS still hasn't implemented Core2 and adding too many things there or moving existing things there also is a bother for them because they can't follow what we're doing. While we're transitioning and until they have implemented Core2, they'd prefer that we don't change too many things 

Core2 should be implemented soon by iOS

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferAlertDialog.kt
```diff
@@ -31,15 +31,36 @@ import com.infomaniak.swisstransfer.ui.theme.Margin
 import com.infomaniak.swisstransfer.ui.theme.SwissTransferTheme
 import com.infomaniak.core2.R as RCore2
 
+object SwissTransferAlertDialogDefaults {
```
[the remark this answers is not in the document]
LunarX: ✌️

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferButton.kt
LunarX: I don't know what you mean by "action button" but I wanted to explicit that this button is only going to be a basic "text" with no fancy background

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
LunarX: This was needed at some point but is not anymore actually, I removed it entirely

tevincent (approved):

PR #256, opened 2024-12-17
fix: Reset transfer status when we received an error from the API

tevincent:


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -73,7 +75,15 @@ fun ImportFilesScreen(
...
 
-    HandleSendActionResult({ sendActionResult }, { selectedTransferType }, navigateToUploadProgress)
+    val snackbarHostState = remember { SnackbarHostState() }
+
+    HandleSendActionResult(
+        snackbarHostState = snackbarHostState,
+        getSendActionResult = { sendActionResult },
+        transferType = { selectedTransferType },
+        navigateToUploadProgress = navigateToUploadProgress,
+        resetSendActionResult = { importFilesViewModel.resetSendActionResult() },
```
KevinBoulongne, line 85:
```suggestion
        resetSendActionResult = importFilesViewModel::resetSendActionResult,
```

KevinBoulongne (changes requested):

PR #264, opened 2024-12-18
chore: Show error message if the env.properties file is missing

LouisCAD:


app/build.gradle.kts
```diff
@@ -15,9 +15,12 @@ val sharedCompileSdk: Int by rootProject.extra
 val sharedMinSdk: Int by rootProject.extra
 val sharedJavaVersion: JavaVersion by rootProject.extra
 
-val envProperties = loadProperties("env.properties")
-val sentryAuthToken = envProperties.getProperty("sentryAuthToken").takeIf { it.isNotEmpty() }
-    ?: error("The `sentryAuthToken` property in `env.properties` must be defined and not empty (see `env.example.properties`).")
+val envProperties = rootProject.file("env.properties").takeIf { it.exists() }?.let { file ->
+    Properties().also { it.load(file.reader()) }
+}
```
tevincent, line 20: We can also display an error if the file does not exist at all. Because right now, we can not differentiate between not having the file and not having the `sentryAuthToken` property.

Also, with your modification, the PR title should be `chore: Show error message if the sentryAuthToken property is missing
` right ?

tevincent (changes requested):

app/build.gradle.kts
LouisCAD: Before this commit, our error message was only shown if the `env.properties` file existed, but was lacking the `sentryAuthToken` property.

Now, we show the error message when the file doesn't exist too, instead of letting the `loadProperties(…)` function crash with no indication, so the PR title is accurate.

I think having a separate error message is overkill, this is good enough to get which file needs to be created and what is expected inside IMO:

> The `sentryAuthToken` property in `env.properties` must be specified (see `env.example.properties`)

sirambd (approved):

tevincent (approved):

PR #265, opened 2024-12-18
chore: Remove reflect for toDestination

sirambd:


KevinBoulongne (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/components/MainScaffold.kt
```diff
@@ -17,7 +17,6 @@
  */
 package com.infomaniak.swisstransfer.ui.screen.main.components
 
-import android.content.Intent
```
KevinBoulongne, line 20: Test CI.

KevinBoulongne (changes requested):

PR #268, opened 2024-12-18
feat: Implementing the "Edit transfer" feature

KevinBoulongne:
Depends on #257


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/upload/UploadProgressViewModel.kt
```diff
@@ -91,6 +91,10 @@ class UploadProgressViewModel @Inject constructor(
         }
     }
 
+    fun removeAllUploadSession() {
+        viewModelScope.launch(ioDispatcher) { uploadManager.removeAllUploadSession() }
+    }
```
KevinBoulongne, line 96: This works.
But I'm not certain it's the best place to do it.
Any insight is appreciated.

PR #275, opened 2024-12-19
chore: Add Proxyman

KevinBoulongne:


app/src/main/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,33 @@
...
+<network-security-config>
+    <debug-overrides>
+        <trust-anchors>
+            <!-- Trust user added CAs while debuggable only -->
+            <certificates src="user" />
+            <certificates src="system" />
+        </trust-anchors>
+    </debug-overrides>
+
+    <base-config cleartextTrafficPermitted="true">
```
FabianDevel, line 27: Are we really forced to allow clearTextTraffic even if it's not in debug ?

FabianDevel (changes requested):

FabianDevel (approved):

PR #277, opened 2024-12-19
feat: Split app integrity logic from send transfer logic

LunarX:
I'll later need to call two different logics with an integrity token so I refactored the code to be able to easily reuse the integrity token for any purpose

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/TransferSendManager.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/TransferSendManager.kt
@@ -111,7 +111,7 @@ class TransferSendManager @Inject constructor(
                 )
             }
 
-            attestationToken?.let { onSuccess(it) } ?: onRefused.invoke()
+            attestationToken?.let(onSuccess) ?: onRefused()
         }.onFailure {
             onFailure.invoke(it)
         }
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/TransferSendManager.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/TransferSendManager.kt
@@ -57,12 +57,8 @@ class TransferSendManager @Inject constructor(
         _integrityCheckResult.value = AppIntegrityResult.Ongoing
 
         withIntegrityToken(
-            onSuccess = { attestationToken ->
-                sendTransfer(newUploadSession, attestationToken)
-            },
-            onRefused = {
-                _integrityCheckResult.value = AppIntegrityResult.Fail
-            },
+            onSuccess = { attestationToken -> sendTransfer(newUploadSession, attestationToken) },
+            onRefused = { _integrityCheckResult.value = AppIntegrityResult.Fail },
             onFailure = { exception ->
                 if (exception !is CancellationException) {
                     SentryLog.e(TAG, "Integrity token received an exception", exception)
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/TransferSendManager.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/TransferSendManager.kt
@@ -29,6 +29,7 @@ import com.infomaniak.swisstransfer.ui.screen.newtransfer.ImportFilesViewModel.A
 import com.infomaniak.swisstransfer.ui.screen.newtransfer.ImportFilesViewModel.SendActionResult
 import com.infomaniak.swisstransfer.workers.UploadWorker
 import dagger.hilt.android.scopes.ViewModelScoped
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.coroutineScope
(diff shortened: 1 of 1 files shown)


FabianDevel (changes requested):
First comments
We need to check the code together, this does not work

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/TransferSendManager.kt
```diff
@@ -53,58 +53,100 @@ class TransferSendManager @Inject constructor(
...
+
+            attestationToken?.let { onSuccess(it) } ?: onRefused.invoke()
+        }.onFailure {
+            onFailure.invoke(it)
+        }
+    }
+
+    private suspend fun requestAppIntegrityToken(appIntegrityManager: AppIntegrityManager): String? = coroutineScope {
+        var attestationToken: String? = null
+
```
FabianDevel, line 122: This won't work because the attestation won't be assign with the good value as it's inside the coroutineScope, so it won't wait until the `launch`completion. 
We need to place the variable outside of the scope in order to execute the launch before the assignation. 

```suggestion
    private suspend fun requestAppIntegrityToken(appIntegrityManager: AppIntegrityManager): String? {
        var attestationToken: String? = null

        coroutineScope {
```

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/TransferSendManager.kt
```diff
@@ -53,58 +53,102 @@ class TransferSendManager @Inject constructor(
...
-            }.onFailure { exception ->
-                SentryLog.e(TAG, "Failed to start the upload", exception)
+
+        withIntegrityToken(
+            onSuccess = { attestationToken ->
+                sendTransfer(newUploadSession, attestationToken)
+            },
+            onRefused = {
+                _integrityCheckResult.value = AppIntegrityResult.Fail
+            },
```
FabianDevel, line 65:
```suggestion
            onSuccess = { attestationToken -> sendTransfer(newUploadSession, attestationToken) },
            onRefused = {  _integrityCheckResult.value = AppIntegrityResult.Fail },
```

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/TransferSendManager.kt
```diff
@@ -53,58 +53,102 @@ class TransferSendManager @Inject constructor(
...
+            onFailure = { exception ->
+                if (exception !is CancellationException) {
+                    SentryLog.e(TAG, "Integrity token received an exception", exception)
+                } else {
+                    SentryLog.i(TAG, "Integrity token received an exception", exception)
+                }
                 _sendActionResult.update { SendActionResult.Failure }
             }
```
FabianDevel, line 73: This onFailure seems useless as it won't catch exceptions happening inside coroutines, and then sendTransfer has its own runCatching
But better safe than sorry I guess, but at least we could directly put this in the onFailure in the function and not expose both onFailure and onRefused(), while onRefused is already applicated when an exeception has been thrown somewhere during appIntegrity

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/TransferSendManager.kt
```diff
@@ -53,58 +53,102 @@ class TransferSendManager @Inject constructor(
...
-                    Log.i(APP_INTEGRITY_MANAGER_TAG, "getApiIntegrityVerdict: $attestationToken")
-                    _integrityCheckResult.value = AppIntegrityResult.Success
-                    sendTransfer(attestationToken, newUploadSession)
+    private suspend fun requestAppIntegrityToken(appIntegrityManager: AppIntegrityManager): String? {
+        var attestationToken: String? = null
+
+        coroutineScope {
+            appIntegrityManager.requestClassicIntegrityVerdictToken(
+                onSuccess = { token ->
+                    SentryLog.i(APP_INTEGRITY_MANAGER_TAG, "request for app integrity token successful $token")
```
FabianDevel, line 126: I'd rather not put the token on sentry

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/TransferSendManager.kt
LunarX: I agree, I removed it from the sentry altogether

FabianDevel (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/TransferSendManager.kt
FabianDevel: We'll fix this in the next PR that already modify this part

KevinBoulongne (approved):
Ok for me, the `runCatching` issues are already fixed in #278.

PR #278, opened 2024-12-19
feat: Add `resendLastTransfer()` function in TransferSend manager

KevinBoulongne:
Depends on #277

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/TransferSendManager.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/TransferSendManager.kt
@@ -29,7 +29,6 @@ import com.infomaniak.swisstransfer.ui.screen.newtransfer.ImportFilesViewModel.A
 import com.infomaniak.swisstransfer.ui.screen.newtransfer.ImportFilesViewModel.SendActionResult
 import com.infomaniak.swisstransfer.workers.UploadWorker
 import dagger.hilt.android.scopes.ViewModelScoped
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/TransferSendManager.kt
```diff
@@ -46,45 +46,55 @@ class TransferSendManager @Inject constructor(
...
-                attestationToken = attestationToken,
-            )!! // TODO: Handle ContainerErrorsException here
-            uploadWorkerScheduler.scheduleWork(uuid)
-            _sendActionResult.update {
-                val totalSize = importationFilesManager.importedFiles.value.sumOf { it.fileSize }
-                SendActionResult.Success(totalSize)
-            }
+            _integrityCheckResult.value = AppIntegrityResult.Ongoing
+
+            withIntegrityToken(
```
FabianDevel, line 73: we can remove the runCatching inside this function that is useless and thus remove the onFailure callback

FabianDevel: Also the runCatching here at top level could be moved inside onSuccess without effect (appart from the `onRefused` whose execution would fall outside of the runCatching, but we shouldn't but code that could throw in this callback)

The lisibility would be easier, and the onSuccess could be extracted to another function

FabianDevel (changes requested):

sirambd (approved):

FabianDevel (approved):

PR #280, opened 2024-12-20
feat: Merge send transfer states

LunarX:
Was required to use the code in multiple screens

Depends on #278

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
@@ -411,7 +411,7 @@ private fun SendButton(
 }
 
 private fun SendStatus.canEnableButton(): Boolean = when (this) {
-    SendStatus.Default,
+    SendStatus.Initial,
     SendStatus.Refused -> true
     else -> false
 }
@@ -478,7 +478,7 @@ private fun Preview(@PreviewParameter(FileUiListPreviewParameter::class) files: 
             addFiles = {},
             closeActivity = {},
             shouldStartByPromptingUserForFiles = false,
-            sendStatus = { SendStatus.Default },
+            sendStatus = { SendStatus.Initial },
             sendTransfer = {},
         )
     }
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
@@ -154,9 +154,13 @@ private fun HandleSendActionResult(
 
     LaunchedEffect(sendStatus()) {
         when (val actionResult = sendStatus()) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -142,52 +135,38 @@ fun ImportFilesScreen(
...
-            is SendActionResult.Success -> navigateToUploadProgress(transferType(), actionResult.totalSize)
-            is SendActionResult.Failure -> {
-                snackbarHostState.showSnackbar(errorMessage)
+    val context = LocalContext.current
+
+    LaunchedEffect(sendStatus()) {
+        when (val actionResult = sendStatus()) {
+            is SendStatus.Success -> navigateToUploadProgress(transferType(), actionResult.totalSize)
+            is SendStatus.Refused -> {
+                snackbarHostState.showSnackbar(context.getString(R.string.errorAppIntegrity))
```
FabianDevel, line 159: Shouldn't we also reset the status here ?

FabianDevel: Also for success in fact
When we reach one of the 3 final stages, we reset it, that seems logic

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -427,13 +403,19 @@ private fun SendButton(
...
+        showIndeterminateProgress = { sendStatus() == SendStatus.Pending },
+        enabled = { importedFiles().isNotEmpty() && !isImporting && isSenderEmailCorrect && sendStatus().canEnableButton() },
         progress = progress,
         onClick = { sendTransfer() },
     )
 }
 
+private fun SendStatus.canEnableButton(): Boolean = when (this) {
+    SendStatus.Default,
+    SendStatus.Refused -> true
```
FabianDevel, line 415: If the state is Failure, shouldn't the button should be re-enabled too?

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
LunarX: If the state is failure or success, we're navigating elsewhere, so thought it'd be safer to disable the button to avoid clicking on it once more or something

FabianDevel (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
LunarX: Yes I'm afraid we can't avoid resetting the state

PR #282, opened 2024-12-20
feat: Implementing "Edit" & "Retry" features

KevinBoulongne:


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/upload/UploadProgressViewModel.kt
```diff
@@ -91,6 +96,20 @@ class UploadProgressViewModel @Inject constructor(
         }
     }
 
+    fun removeAllUploadSession(onCompletion: () -> Unit) {
+        viewModelScope.launch(ioDispatcher) {
```
LouisCAD, line 100: This is not necessary because there's no blocking I/O done here at all. We can leave the default.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/upload/UploadProgressViewModel.kt
```diff
@@ -91,6 +96,20 @@ class UploadProgressViewModel @Inject constructor(
...
+            uploadManager.removeAllUploadSession()
+            withContext(mainDispatcher) { onCompletion() }
+        }
+    }
+
+    fun resendLastTransfer(onCompletion: () -> Unit) {
+        viewModelScope.launch(ioDispatcher) {
```
LouisCAD, line 107: As just above, this is not necessary because there's no blocking I/O done here at all. We can leave the default.

LouisCAD (approved):
Looks good to me. Added 2 comments on something we can improve in this PR, or later on.

PR #299, opened 2025-01-03
feat: Opt-out of Matomo & Sentry

LunarX:
I store the matomo/sentry opt out state in a data store and read that value to prevent sending matomo/sentry events if the user has opted out.

To do this while centralizing the definition of data store's entries I had to add a few helper methods.

I also removed `NetworkException`s from Sentry the same way we did it on mail.

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/SettingsMatomoViewModel.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/SettingsMatomoViewModel.kt
@@ -21,14 +21,12 @@ import android.content.Context
 import androidx.datastore.preferences.core.edit
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
-import com.infomaniak.swisstransfer.di.IoDispatcher
 import com.infomaniak.swisstransfer.ui.utils.DataManagementPreferences
 import com.infomaniak.swisstransfer.ui.utils.dataManagementDataStore
 import com.infomaniak.swisstransfer.ui.utils.get
 import com.infomaniak.swisstransfer.ui.utils.set
 import dagger.hilt.android.lifecycle.HiltViewModel
 import dagger.hilt.android.qualifiers.ApplicationContext
-import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.stateIn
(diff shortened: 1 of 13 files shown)


gradle/libs.versions.toml
```diff
@@ -8,6 +8,7 @@ composeBom = "2024.12.01"
 constraintlayoutCompose = "1.1.0"
 coreKtx = "1.15.0"
 coreSplashscreen = "1.0.1"
+datastorePreferencesAndroid = "1.1.1"
```
LouisCAD, line 11: We probably don't need the "Android" suffix here. BTW, androidx.datastore is multiplatform.

gradle/libs.versions.toml
```diff
@@ -32,6 +33,7 @@ androidx-adaptive-navigation = { module = "androidx.compose.material3.adaptive:a
 androidx-constraintlayout-compose = { module = "androidx.constraintlayout:constraintlayout-compose", version.ref = "constraintlayoutCompose" }
 androidx-core-ktx = { group = "androidx.core", name = "core-ktx", version.ref = "coreKtx" }
 androidx-core-splashscreen = { module = "androidx.core:core-splashscreen", version.ref = "coreSplashscreen" }
+androidx-datastore-preferences = { group = "androidx.datastore", name = "datastore-preferences", version.ref = "datastorePreferencesAndroid" }
```
LouisCAD, line 36: Same as above, the "Android" suffix in the version ref is unnecessary IMO.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/SettingsMatomoViewModel.kt
```diff
@@ -0,0 +1,54 @@
...
+class SettingsMatomoViewModel @Inject constructor(
+    @ApplicationContext private val appContext: Context,
+    @IoDispatcher private val ioDispatcher: CoroutineDispatcher,
+) : ViewModel() {
+    val isMatomoAuthorized = appContext.dataManagementDataStore.data
+        .map { it[DataManagementPreferences.IsMatomoAuthorized] }
+        .stateIn(viewModelScope, SharingStarted.Lazily, false)
+
+    fun setMatomoAuthorization(isAuthorized: Boolean) {
+        viewModelScope.launch(ioDispatcher) {
```
LouisCAD, line 48: We don't need the ioDispatcher, the DataStore doesn't do any blocking I/O on the threads its suspend functions are called on. Just use `launch { … }`

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/SettingsSentryViewModel.kt
```diff
@@ -0,0 +1,54 @@
...
+class SettingsSentryViewModel @Inject constructor(
+    @ApplicationContext private val appContext: Context,
+    @IoDispatcher private val ioDispatcher: CoroutineDispatcher,
+) : ViewModel() {
+    val isSentryAuthorized = appContext.dataManagementDataStore.data
+        .map { it[DataManagementPreferences.IsSentryAuthorized] }
+        .stateIn(viewModelScope, SharingStarted.Lazily, false)
+
+    fun setSentryAuthorization(isAuthorized: Boolean) {
+        viewModelScope.launch(ioDispatcher) {
```
LouisCAD, line 48: Like above, ioDispatcher isn't needed.

LouisCAD (approved):

LouisCAD (approved):
LGTM

PR #303, opened 2025-01-07
refactor: Rename `TransferTypeUi` enum before using its name

KevinBoulongne:


LouisCAD (approved):
I assume those enums are not persisted anywhere in the app data.
Looks good to me.

PR #324, opened 2025-01-09
feat: Add transfer sections with sticky headers

LunarX:
This PR includes some unit tests to make sure the grouping logic is correct


app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItemList.kt
```diff
@@ -53,37 +60,54 @@ fun TransferItemList(
...
-            count = getTransfers().count(),
-            key = { getTransfers()[it].uuid },
-            contentType = { getTransfers()[it] },
-            itemContent = {
-                val transfer = getTransfers()[it]
-                SwipeToDismissComponent(
-                    contentShape = itemShape,
-                    onSwiped = { onSwiped(transfer.uuid) },
+        getTransfers().forEach { (section, transfers) ->
+            stickyHeader {
```
sirambd, line 70: Missing key

sirambd (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItemList.kt
```diff
@@ -53,37 +60,54 @@ fun TransferItemList(
...
+                Box(
+                    modifier = Modifier
+                        .fillParentMaxWidth()
+                        .background(SwissTransferTheme.materialColors.background)
```
sirambd, line 74: This background is really needed ?

sirambd (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transfers/TransfersGroupingManager.kt
```diff
@@ -0,0 +1,100 @@
...
+                else -> TransferSection.ByMonth(getMonthNameFromEpoch(transfer.createdDateTimestamp))
+            }
+        }
+    }
+
+    sealed class TransferSection(val title: Context.() -> String) {
+        data class ByMonth(val monthName: String) : TransferSection({ monthName })
+    }
+
+    sealed class SpecificSection(
```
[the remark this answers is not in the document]
LunarX: `SpecificSection` regroups all of the types that need to define a "contains" expression. `ByMonth` is the only one which doesn't need it. I needed a common interface to define my `isIn` method but it makes no sense to define a contains method for `ByMonth` so I needed it to be defined elsewhere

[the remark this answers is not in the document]
LunarX: But maybe renaming `SpecificSection` to something like `WellDefinedSection` or something else could still help?

app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItemList.kt
LunarX: Yes, this way the sticky header is a whole width background-colored rectangle with text on it. Else the sticky header is only text on a transparent background

sirambd (approved):

PR #341, opened 2025-01-14
feat: Prevent file picker from opening when backing after a canceled transfer

LunarX:
Starting an upload and canceling it brings you back to the ImportFilesScreen. When doing so, the code would trigger and open the file picker again which we don't want.

Using a channel, we make sure the file picker event is only consumed once


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/ImportFilesScreen.kt
```diff
@@ -165,9 +167,9 @@ fun ImportFilesScreen(
 }
 
 @Composable
-private fun HandleStartupFilePick(shouldPickFilesOnStartup: Boolean, pickFiles: () -> Unit) {
-    LaunchedEffect(shouldPickFilesOnStartup) {
-        if (shouldPickFilesOnStartup) pickFiles()
+private fun HandleStartupFilePick(openFilePickerEvent: ReceiveChannel<Unit>, pickFiles: () -> Unit) {
+    LaunchedEffect(Unit) {
```
LouisCAD, line 171: Even though the current code isn't broken by that, it's a good practice to have the parameters used in the `LaunchedEffect` passed as a key, `pickFiles`, in this situation.
```suggestion
    LaunchedEffect(pickFiles) {
```

LouisCAD (approved):
Good.

PR #342, opened 2025-01-14
fix: Correctly manage the app tasks when backing depending on isTaskRoot

LunarX:
The app task was empty but still there when sharing a file to the app when the app was never launched before

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/NewTransferScreen.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/NewTransferScreen.kt
@@ -17,6 +17,7 @@
  */
 package com.infomaniak.swisstransfer.ui.screen.newtransfer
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/NewTransferScreen.kt
```diff
@@ -36,6 +37,10 @@ fun NewTransferScreen(closeActivity: () -> Unit) {
     val navController = rememberNavController()
     var displayConfirmationDialog by rememberSaveable { mutableStateOf(false) }
 
+    BackHandler(!displayConfirmationDialog) {
+        closeActivity()
+    }
```
LunarX, line 42: Check if currentDestination is the import screen or else this will trigger for any screen that doesn't override the BackHandler

sirambd:
It's doesn't work if NewTransferActivity is launched before sharing files

PR #345, opened 2025-01-15
chore: Update urlUserReportAndroid string key into urlUserReport

LunarX:
The string userfeedback thread will be the same for Android and iOS unlike the mail ones. It is expected that the URLs that currently link to the web userfeedback threads will be updated in the future to link to threads specific to mobile instead

### DIFF
--- a/app/src/main/res/values-fr/strings.xml
+++ b/app/src/main/res/values-fr/strings.xml
@@ -180,7 +180,7 @@
     <string name="uploadSuccessLinkTitle">Ton transfert est prêt !</string>
     <string name="uploadSuccessQrTitle">Ton QR code est prêt !</string>
     <string name="urlAbout">https://www.infomaniak.com/fr/a-propos</string>
-    <string name="urlUserReportAndroid">https://feedback.userreport.com/1c462a20-7559-415e-a6e0-4b624dc38877/#ideas/popular</string>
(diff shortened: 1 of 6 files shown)


app/src/main/res/values-fr/strings.xml
```diff
@@ -180,7 +180,7 @@
     <string name="uploadSuccessLinkTitle">Ton transfert est prêt !</string>
     <string name="uploadSuccessQrTitle">Ton QR code est prêt !</string>
     <string name="urlAbout">https://www.infomaniak.com/fr/a-propos</string>
-    <string name="urlUserReportAndroid">https://feedback.userreport.com/1c462a20-7559-415e-a6e0-4b624dc38877/#ideas/popular</string>
+    <string name="urlUserReport">https://feedback.userreport.com/1c462a20-7559-415e-a6e0-4b624dc38877/#ideas/popular</string>
```
KevinBoulongne, line 183: Same than the spanish one. Is it on purpose ?

KevinBoulongne (changes requested):

app/src/main/res/values-fr/strings.xml
KevinBoulongne: Looks like it's the 🇪🇸 one that is wrong.
The link is the one for the 🇫🇷 language.

LunarX: Nice catch, I updated the strings it's fixed

KevinBoulongne (approved):

PR #343, opened 2025-01-14
feat: Show transfer download status live

LouisCAD:
- Follow progress and display it to the user
- Connect back to the system's DownloadManager
- Allow cancelling the download
- Allow opening the download
- Tell the user if an error happens while downloading
- Automatically detect if the file is deleted externally

PR #346, opened 2025-01-15
feat: Per file download + open

LouisCAD:
Depends on #343


LouisCAD:
We probably want to show a share icon once the file is downloaded, instead of the checkmark.

We might also want to implement long press to delete (with confirmation) to allow the user to clean things up directly from SwissTransfer.

sirambd (approved):

PR #347, opened 2025-01-16
feat: Open deeplink that you sent the right way

FabianDevel:


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/DeeplinkViewModel.kt
```diff
@@ -0,0 +1,50 @@
...
+import dagger.hilt.android.lifecycle.HiltViewModel
+import javax.inject.Inject
+
+@HiltViewModel
+class DeeplinkViewModel @Inject constructor(
+    private val savedStateHandle: SavedStateHandle,
+    private val transferManager: TransferManager,
+) : ViewModel() {
+
+    val isDeepLinkConsumed = savedStateHandle.getStateFlow(IS_DEEP_LINK_CONSUMED_KEY, false)
```
tevincent, line 33:
```suggestion
    val isDeeplinkConsumed = savedStateHandle.getStateFlow(IS_DEEP_LINK_CONSUMED_KEY, false)
```

You also have other `DeepLink` in that file.

tevincent (changes requested):

tevincent (approved):

PR #350, opened 2025-01-16
fix: Do not display file picker when the activity is killed and recreated if it's not needed

LunarX:
When the activity is killed and recreated, remember to not handle `NewTransferOpenManager.Reason.Other` thanks to `isFirstViewModelCreation`


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/ImportFilesViewModel.kt
```diff
@@ -18,6 +18,7 @@
 package com.infomaniak.swisstransfer.ui.screen.newtransfer
 
 import android.net.Uri
+import android.util.Log
```
tevincent, line 21:
```suggestion
```

tevincent (changes requested):

tevincent (approved):

PR #357, opened 2025-01-17
fix: Handle NewTransferActivity task correctly with the sharing of files

LunarX:
Clear task when NewTransferActivity is the root of the task and make sure that the NewTransferActivity can't be stacked multiple times if files are shared to it

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -110,7 +110,7 @@
 
         <activity
             android:name=".ui.NewTransferActivity"
-            android:launchMode="singleTask" />
+            android:launchMode="standard" />
     </application>
 
 </manifest>
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -108,9 +108,9 @@
             android:name=".ui.MainActivity"
             android:launchMode="singleTask" />
 
-        <activity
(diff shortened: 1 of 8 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -108,9 +108,9 @@
             android:name=".ui.MainActivity"
             android:launchMode="singleTask" />
 
-        <activity
-            android:name=".ui.NewTransferActivity"
-            android:launchMode="singleTask" />
```
KevinBoulongne, line 113: Commenting only to block this PR while I'm checking if this doesn't break notifications.

KevinBoulongne (changes requested):

sirambd (approved):

tevincent (approved):

app/src/main/AndroidManifest.xml
KevinBoulongne: As seen with @LunarX, it breaks notifications.
But it should be easy to fix, we just need to decide what behavior we prefer.

KevinBoulongne: Finally, it was easy to fix, but not easy to find the correct fix.
A lot of circonvolution were involved.

KevinBoulongne (approved):

PR #358, opened 2025-01-17
feat: Add build variant to switch easily from prod to preprod

tevincent:


app/build.gradle.kts
```diff
@@ -26,6 +26,10 @@ android {
     namespace = "com.infomaniak.swisstransfer"
     compileSdk = sharedCompileSdk
 
+
```
KevinBoulongne, line 29:
```suggestion
```

app/build.gradle.kts
```diff
@@ -54,11 +54,32 @@ android {
         release {
             isMinifyEnabled = true
             isShrinkResources = true
+
```
KevinBoulongne, line 57:
```suggestion
```

KevinBoulongne (changes requested):

app/build.gradle.kts
```diff
@@ -54,11 +54,32 @@ android {
...
+            applicationIdSuffix = ".preprod"
+
+            buildConfigField("String", "BASE_URL", "\"https://$preprodHost\"")
+        }
+        create("prod") {
+            dimension = "env"
+
+            buildConfigField("String", "BASE_URL", "\"https://$prodHost\"")
+        }
```
KevinBoulongne, line 77: We should change the app icon background color for each of these 4 possibilities.

KevinBoulongne (changes requested):

app/build.gradle.kts
```diff
@@ -68,15 +94,25 @@ android {
         sourceCompatibility = sharedJavaVersion
         targetCompatibility = sharedJavaVersion
     }
+
+    signingConfigs {
+        getByName("debug") {
+            storeFile = file("./debug.keystore")
+        }
+    }
```
LouisCAD, line 102: I think you forgot to use it in the buildType. You need to reference it above.

_BTW, we might want to put the keystore file at the root of the project, and use `rootProject.file(…)` in case we want to add another application module in this project in the future._

LouisCAD (changes requested):

KevinBoulongne (approved):

LouisCAD (approved):
We might leverage drawable references in product flavor resources, so actual assets are all in the main sourceSet, allowing us to see the rendered versions without having to switch the buildVariant in the IDE.

PR #365, opened 2025-01-17
feat: Replace static illustration with lottie

tevincent:


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/onboarding/OnboardingScreen.kt
```diff
@@ -68,7 +70,11 @@ fun OnboardingScreen(goToMainActivity: () -> Unit) {
     }
 
     val onboardingPages = buildList {
-        Page.entries.forEach { page -> add(page.toOnboardingPage(isHighlighted)) }
+        Page.entries.forEachIndexed { index, page ->
+            val isPlayingLottie by remember { derivedStateOf { pagerState.currentPage == index } }
+
+            add(page.toOnboardingPage(isHighlighted, isPlayingLottie = { isPlayingLottie }))
```
LunarX, line 76: Maybe we should put the isPlayingLottie logic inside toOnboardingPage as it's part of the creating the OnboardingPage UI

LunarX (changes requested):

PR #373, opened 2025-01-21
feat: Refresh transfer list

FabianDevel:
Update all the transfers att the start of the mainActivity, but at most once per 15min


LouisCAD (approved):
Perfect!

PR #375, opened 2025-01-21
feat: Sync download status across UI for single-file transfer

LouisCAD:


LouisCAD:
Branch had wrong name, superseded by #376

PR #376, opened 2025-01-21
feat: Sync download status across UI for single-file transfer

LouisCAD:


FabianDevel (approved):
LGTM, but we should remove getDownloadIdFromStorage from the KMP module

PR #396, opened 2025-01-29
fix: Keep same display logic in transfer details screen

NicolasBourdin88:
Before : 
Screenshot_1738146795
After : 
Screenshot_1738146624


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/filesdetails/NewTransferFilesDetailsScreen.kt
```diff
@@ -79,7 +79,7 @@ private fun NewTransferFilesDetailsScreen(
     navigateBack: (() -> Unit),
 ) {
     val snackbarHostState = remember { SnackbarHostState() }
-    Scaffold(
+    SinglePaneScaffold(
```
LunarX, line 82: It was quicker to fix than I thought. Is this the only screen that's missing this horizontal maximum width constraint?

LunarX (approved):

PR #398, opened 2025-01-30
feat: Check write permission for API <= 28

KevinBoulongne:
Depends on #397

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/utils/PermissionsUtils.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/utils/PermissionsUtils.kt
@@ -23,7 +23,7 @@ import com.google.accompanist.permissions.PermissionState
 import com.google.accompanist.permissions.isGranted
 
 @OptIn(ExperimentalPermissionsApi::class)
-fun PermissionState?.checkWritePermission(action: () -> Unit) {
+fun PermissionState?.checkPermission(action: () -> Unit) {
     if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.P && this?.status?.isGranted != true) {
         this?.launchPermissionRequest()
     } else {
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/utils/PermissionsUtils.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/utils/PermissionsUtils.kt
@@ -1,0 +1,32 @@
+/*
+ * Infomaniak SwissTransfer - Android
+ * Copyright (C) 2025 Infomaniak Network SA
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/utils/PermissionsUtils.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.google.accompanist.permissions.isGranted
+
+@OptIn(ExperimentalPermissionsApi::class)
+fun PermissionState?.checkWritePermission(action: () -> Unit) {
+    if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.P && this?.status?.isGranted != true) {
+        this?.launchPermissionRequest()
+    } else {
+        action()
+    }
+}
```
LouisCAD, line 32: The function name is not super accurate to me.
However, there's a UX problem that would justify doing something else, and finding another name.

Here's the UX problem:
1. The user tries to download on API <= 28
2. The app doesn't have the write permission yet
3. It launches the permission request
4. The user grants it
5. **Download doesn't start**

I think it'd be better to remember the user intent and start downloading automatically after the permission is granted.

To do that, we can leverage Compose's `snapshotFlow { … }` that will emit a new Flow value whenever some state read in its lambda changes.

That means we can turn `PermissionState` into a `Flow<Boolean>` that emits true once the permission is granted.

We can then make such a function:
```kotlin
@Composable
PermissionState?.guardedCallback(callback: () -> Unit): () -> Unit
```

And this function can then be fed directly to `onClick`:

```kotlin
TopAppBarButtons.Download(
    enabled = downloadRequest.isAwaitingCall,
    onClick = writeExternalStoragePermissionState.guardedCallback { downloadRequest() },
)
```

I'd be happy to pair-program this with you!

LouisCAD (approved):
Looks good enough to me, but if you have some time, we can do a bit better, UX-wise.

app/src/main/java/com/infomaniak/swisstransfer/ui/utils/PermissionsUtils.kt
KevinBoulongne: Indeed, the name is oxbow code and should be changed.

The user intent feedback is interesting, I also think we want that.
Let's talk about it on Monday !

KevinBoulongne: The name has been updated.

I'm merging this PR, and we'll do another one next week to handle your suggestion.

app/src/main/java/com/infomaniak/swisstransfer/ui/utils/PermissionsUtils.kt
```diff
@@ -0,0 +1,32 @@
...
+package com.infomaniak.swisstransfer.ui.utils
+
+import android.os.Build
+import com.google.accompanist.permissions.ExperimentalPermissionsApi
+import com.google.accompanist.permissions.PermissionState
+import com.google.accompanist.permissions.isGranted
+
+@OptIn(ExperimentalPermissionsApi::class)
+fun PermissionState?.checkPermission(action: () -> Unit) {
+    if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.P && this?.status?.isGranted != true) {
```
LouisCAD, line 27: @KevinBoulongne I forgot to mention that this condition is dangerous, because this function might be used in other parts of the app, and then, it'd break for API levels greater than "P" (BTW, can we use literal integers for that, please 🙏 )

app/src/main/java/com/infomaniak/swisstransfer/ui/utils/PermissionsUtils.kt
LouisCAD: Addressed in #402

PR #403, opened 2025-02-03
fix: Fix transfer change not taken into account

LouisCAD:
This is certainly what caused the download button to not update on tablets (mater/details UI) when selecting another transfer in the list.


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/TransferDetailsScreen.kt
```diff
@@ -165,7 +166,7 @@ private fun TransferDetailsScreen(
     }
     val transferFlow = remember { snapshotFlow { getTransfer() } }
     LaunchedEffect(Unit) {
-        transferFlow.collect { transfer ->
+        transferFlow.collectLatest { transfer ->
```
KevinBoulongne, line 169: This doesn't fix the issue

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #406, opened 2025-02-04
chore: Declare handled config changes

LouisCAD:
This is safe to do since all the APIs we use in Compose should already handle configChanges.

I made sure that `Context` objects are used as a remember keys when applicable.

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/FilePreview.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/FilePreview.kt
@@ -73,7 +73,7 @@ private val videoFrameDecoderFactory = VideoFrameDecoder.Factory()
 @Composable
 private fun FileThumbnail(uri: String, onError: () -> Unit) {
     val context = LocalContext.current
-    val imageRequest = remember(uri) {
+    val imageRequest = remember(uri, context) {
         ImageRequest.Builder(context)
             .data(uri)
             .decoderFactory(videoFrameDecoderFactory)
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/components/FilePreview.kt
```diff
@@ -73,7 +73,7 @@ private val videoFrameDecoderFactory = VideoFrameDecoder.Factory()
 @Composable
 private fun FileThumbnail(uri: String, onError: () -> Unit) {
     val context = LocalContext.current
-    val imageRequest = remember(uri) {
+    val imageRequest = remember(uri, context) {
```
tevincent, line 76: Maybe we can put the context as a key always first ?

tevincent (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/components/FilePreview.kt
LouisCAD: It doesn't matter at all, what's important is that the keys that can change are there. It's just like `if (a || b)` vs `if (b || a)`, with `a` and `b` being local variables: no perceptible difference for the end user.

I think we should focus on rules that matter, and write these down in our knowledge base or another appropriate place, rather than inferring undocumented ones every time we see a possible pattern, and slowing down ourselves with extra feedback loops because of that.

sirambd (approved):

tevincent (approved):

PR #412, opened 2025-02-12
feat: Add generation of thumbnails for videos and images

tevincent:


gradle/libs.versions.toml
```diff
@@ -19,7 +19,7 @@ navigation = "2.8.5"
 qrose = "1.0.1"
 recaptcha = "18.6.1"
 sentry = "4.12.0"
-swisstransfer = "1.0.5"
+swisstransfer = "1.0.6-vte"
```
tevincent, line 22: This will be removed when we'll make a relase of KMP.

KevinBoulongne (approved):

PR #421, opened 2025-02-19
chore: Bump libs

JorisBodin:


sirambd (changes requested):
There are a lot of changes in the gradlew script file, so we'll have to test to make sure everything works like the old code, and is there a reason for these changes?

sirambd (approved):

PR #428, opened 2025-02-24
refactor: Rename ExpiredFetchTransferException into ExpiredDateFetchTransferException

KevinBoulongne:


LouisCAD (changes requested):
The commit and PR title should say "replace xx with yy" rather than "rename". The renaming already took place in the KMP library.

LouisCAD (approved):
Thanks!

PR #446, opened 2025-02-28
fix: Don't send sentry for CancellationException

sirambd:


app/src/main/java/com/infomaniak/swisstransfer/ui/MainApplication.kt
```diff
@@ -92,12 +89,13 @@ class MainApplication : Application(), Configuration.Provider {
...
             // Register the callback as an option
             options.beforeSend = SentryOptions.BeforeSendCallback { event: SentryEvent, _: Any? ->
+                if (event.throwable is CancellationException) return@BeforeSendCallback null
                 val isNetworkException = event.exceptions?.any { it.type == "ApiController\$NetworkException" } ?: false
                 /**
                  * Reasons to discard Sentry events :
                  * - Application is in Debug mode
                  * - User deactivated Sentry tracking in DataManagement settings
-                 * - The exception was a NetworkException, and we don't want to send them to Sentry
+                 * - The exception was a NetworkException or [CancellationException], and we don't want to send them to Sentry
```
LunarX, line 98: This block of comment kind of applies to the line below it the was it's written right now. But the CancellationException is handled before this line. Either we should write the NetworkException like the cancel exception or we should put the comment above everything or something

LunarX (changes requested):

LunarX (approved):

PR #450, opened 2025-03-05
feat: Matomo basic

FabianDevel:


app/src/main/java/com/infomaniak/swisstransfer/ui/navigation/NavigationDestination.kt
```diff
@@ -110,48 +111,96 @@ sealed class MainNavigation : NavigationDestination() {
  * Sealed class representing the navigation arguments for the new transfer flow.
  */
 @Serializable
-sealed class NewTransferNavigation : NavigationDestination() {
+sealed class NewTransferNavigation(override val matomoValue: String) : NavigationDestination() {
 
     @Serializable
-    data object ImportFilesDestination : NewTransferNavigation()
+    data object ImportFilesDestination : NewTransferNavigation("ImportFileView")
```
LunarX, line 117: This destination should soon change after @LouisCAD 's PR. Would we want to already use a name that matches the new screen name so we don't send incoherent values to matomo?

app/src/main/java/com/infomaniak/swisstransfer/ui/navigation/NavigationDestination.kt
```diff
@@ -110,48 +111,96 @@ sealed class MainNavigation : NavigationDestination() {
...
-    ) : NewTransferNavigation()
+    ) : NewTransferNavigation("UploadSuccessView")
 
     @Serializable
     data class UploadErrorDestination(
         val transferType: TransferTypeUi,
         val totalSize: Long,
         val authorEmail: String?,
-    ) : NewTransferNavigation()
+    ) : NewTransferNavigation("UploadErrorView")
```
LunarX, line 141: Maybe this will change as well with @LouisCAD 's PR. What will happen to this tracking then?

app/src/main/java/com/infomaniak/swisstransfer/ui/navigation/NavigationDestination.kt
LunarX, line 1: What about naming all values with the "Screen" sufix instead of "View"? It could make more sens as we're really tracking screens maybe?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/TransferTypeButtons.kt
LunarX, line 1: Do we have the shared document that lists all the event names?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/MainScreen.kt
```diff
@@ -36,7 +36,9 @@ fun MainScreen(deeplinkTransferDirection: TransferDirection? = null) {
     val hideBottomBar = rememberSaveable { mutableStateOf(false) }
 
     val currentDestination by remember(navBackStackEntry) {
-        derivedStateOf { navBackStackEntry?.toMainDestination() ?: MainNavigation.startDestination }
+        val destination = navBackStackEntry?.toMainDestination()
+        destination?.trackScreen()
+        derivedStateOf { destination ?: MainNavigation.startDestination }
```
LunarX, line 41: Is this better than a LaunchEffect?

LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/navigation/NavigationDestination.kt
FabianDevel: Now that we have the need to name each destination, we should align with iOS, that's why we should use this name, and not the current nor future class's name

app/src/main/java/com/infomaniak/swisstransfer/ui/navigation/NavigationDestination.kt
FabianDevel: Same as above, we want to keep matomoValues agnostic from our class name to allow to be iso with iOS

app/src/main/java/com/infomaniak/swisstransfer/ui/navigation/NavigationDestination.kt
FabianDevel: We need the "View" value to align with iOS (they already have data on matomo)

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/importfiles/components/TransferTypeButtons.kt
FabianDevel: Yes we do

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/MainScreen.kt
FabianDevel: Well I didn't want to divide the navigation logic at first
But apart from that, I don't really think this is better than a launchedEffect.

The advantage of the Launched effect is that it's not as "hidden" as it is here

I'm pretty neutral tbh

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transfers/TransfersScreenWrapper.kt
```diff
@@ -140,6 +141,7 @@ private fun ThreePaneScaffoldNavigator<DestinationContent>.navigateToDetails(
     direction: TransferDirection,
     transferUuid: String,
 ) {
+    MatomoSwissTransfer.trackScreen("${direction.name.lowercase()}TransferDetailView")
```
KevinBoulongne, line 144: Maybe we should add a comment to the `TransferDirection` enum ?
Something like _"// Don't rename these, as it will break Matomo"_.

Or maybe even better, add a `matomoValue` parameter to this enum ?

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/MainScreen.kt
tevincent: IMO, it's better to have everything in the same place. The only reason to put the `Matomo` logics elsewhere for me is if it's a very complex logic. But here, it's only a few lines so ...

tevincent (approved):

app/src/main/java/com/infomaniak/swisstransfer/ui/navigation/NavigationDestination.kt
FabianDevel: After talking about it in real, the problem is not the fact that he destination will change, it's that some destination will totally be erased

KevinBoulongne (approved):

LunarX (changes requested):
Maybe using a "MatomoScreen" enum to track screen instead of defining strings everywhere would be better. This way all screens that are the same type of objects are defined together in a single place

LunarX (approved):

PR #452, opened 2025-03-18
fix: Improve PickFilesScreen UX

FabianDevel:


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/pickfiles/components/EmailAddressTextField.kt
LunarX, line 1: We could check if we can use Modifier.imePaddings() to show the send button on the screen at all time and easily navigate this screen. If it doesn't work, this is already better than what we had previously

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/pickfiles/components/EmailAddressTextField.kt
```diff
@@ -72,7 +73,13 @@ fun EmailAddressTextField(
     val isImeVisible = WindowInsets.isImeVisible
     LaunchedEffect(isImeVisible) { if (!isImeVisible) state.addRecipientAddress() }
 
-    val keyboardActions = KeyboardActions(onDone = { state.addRecipientAddress() })
+    val keyboardController = LocalSoftwareKeyboardController.current
+    val keyboardActions = KeyboardActions(onDone = {
+        // Using the default onDone behaviours and waiting for the ime launchedEffect to validate the recipient make a small
+        // visible lag to the email transformation, so we enforce the addRecipientAddress here before closing the keyboard
+        state.addRecipientAddress()
```
LunarX, line 80: If we call `addRecipientAddress()` here, then the LaunchedEffect will also trigger later on, but since we add values to a `Set`, it won't cause any problem. Is there no cleaner way of handling the addition of a new recipient?

LunarX (changes requested):

LunarX (approved):

LunarX (approved):

PR #455, opened 2025-03-20
fix: Handle unable to create directory for downloads

LouisCAD:


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/TransferDownload.kt
```diff
@@ -196,10 +197,20 @@ private suspend fun buildDownloadRequest(
...
+            mimeType = FileType.guessMimeTypeFromFileName(name),
+            userAgent = userAgent,
+        )
+    }.onFailure {
+        // On some Xiaomi devices, `setDestinationInExternalPublicDir` might need to create
+        // the Download directory, and might fail at it with an `IllegalStateException`.
+        SentryLog.w(TAG, "Failed to create the DownloadManager request", it)
+        // It's very likely that trying again will work, so we don't bother showing the error,
+        // and just let the user try again.
+    }.getOrNull()
```
FabianDevel, line 213: If it ever fails repeatedly, won't it just infinite loop now that it's not thrown because of the `repeatWhileActive` ?
To which extend are you sure about your "very likely" ?

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/TransferDownload.kt
LouisCAD: Shouldn't because the loop waits for user request (i.e. a click on the connected button) on ever iteration.

LouisCAD: See the line just before the call site here: https://github.com/Infomaniak/android-SwissTransfer/pull/455/files#diff-10a0103e82844f01d6274062d5c395aa58240b9cd1bebf6477d449ea842afccaR148-R157

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/TransferDownload.kt
```diff
@@ -196,10 +197,20 @@ private suspend fun buildDownloadRequest(
...
-        userAgent = userAgent,
-    )
+    return runCatching {
+        DownloadManagerUtils.requestFor(
+            url = url,
+            nameWithoutProblematicChars = name,
+            mimeType = FileType.guessMimeTypeFromFileName(name),
+            userAgent = userAgent,
+        )
+    }.onFailure {
```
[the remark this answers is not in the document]
LouisCAD: The content of the `runCatching` block isn't suspending, here, and isn't checking for cancellation either, so it's not needed.

That said, you made me rethink about it, and while checking, I found out that `setDestinationInExternalPublicDir`, which is called by `DownloadManagerUtils.requestFor`, is doing blocking I/O under the hood, so we want to wrap this in the IO dispatcher, meaning that, yes, we will want to add `.cancellable()`.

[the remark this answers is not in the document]
LouisCAD: Updated core and added `.cancellable()`.

FabianDevel (approved):

PR #458, opened 2025-03-26
fix(SettingsScreen): Fix data management selection when on Matomo or Sentry screens

FabianDevel:


LouisCAD (approved):
We could use a `sealed interface` too, but this works.

PR #461, opened 2025-03-26
fix: Recomposition not happening on configuration change in TransferItemList

TommyDL-Infomaniak:


app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItemList.kt
```diff
@@ -73,9 +78,13 @@ fun TransferItemList(
                 Box(
                     modifier = Modifier
                         .fillParentMaxWidth()
-                        .background(SwissTransferTheme.materialColors.background)
+                        .background(stickyHeaderBackground)
                         .animateItem()
                 ) {
+                    // Needed for section.title below, because the context stays the same on configuration changes.
+                    // Reading this makes sure composition happens.
+                    LocalConfiguration.current
```
tevincent, line 86: Honestly, I have the feeling we want to over optimized things where we don't need to. Because if we have to do hacky things like this in order to have everything working as expected, we'll have other issues in other part of the code. IMO, we should remove the configChanges in the manifest instead of that `Solution`.

tevincent (changes requested):

LunarX (changes requested):
There's no use in fixing anything here whatsoever because we have a card asking to fix all of the recompositions by removing all of the manual config changes from the manifest

sirambd (approved):

PR #463, opened 2025-03-31
feat: Per chunk retry for uploads

LouisCAD:


gradle/libs.versions.toml
```diff
@@ -12,13 +12,13 @@ desugarJDK = "2.1.4"
 hiltAndroid = "2.55"
 hiltAndroidx = "1.2.0"
 hiltNavigationCompose = "1.2.0"
-kotlin = "2.0.21"
+kotlin = "2.1.20"
```
LouisCAD, line 15: Needed to use Kotlin-friendly atomics in `TransferUploader`.

LouisCAD:
Here's a patch to force a kind of error: Chunk was uploaded, but network issue on receiving the confirmation


```patch
Subject: [PATCH] Testing: fail 1st chunk
---
Index: app/src/main/java/com/infomaniak/swisstransfer/upload/TransferUploader.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/src/main/java/com/infomaniak/swisstransfer/upload/TransferUploader.kt b/app/src/main/java/com/infomaniak/swisstransfer/upload/TransferUploader.kt
--- a/app/src/main/java/com/infomaniak/swisstransfer/upload/TransferUploader.kt	(revision b188b6b77b859179cdd04ca45b98868f59fa715a)
+++ b/app/src/main/java/com/infomaniak/swisstransfer/upload/TransferUploader.kt	(date 1743405133407)
@@ -38,6 +38,7 @@
 import kotlinx.coroutines.sync.Semaphore
 import splitties.init.appCtx
 import java.io.BufferedInputStream
+import java.io.IOException
 import java.util.concurrent.ArrayBlockingQueue
 import kotlin.concurrent.atomics.*
 
@@ -263,6 +264,10 @@
                             isLastChunk = isLastChunk,
                             data = dataByteArray
                         )
+                        if (failFirstChunkForTesting && chunkIndex == 0) {
+                            failFirstChunkForTesting = false
+                            throw IOException("Imagine it didn't work")
+                        }
                         SentryLog.i(TAG, "Completed uploading chunk #$chunkIndex of $fileUUID")
                         metadata.chunksUploadStatus[chunkIndex] = DefinitelyComplete
                         // Allow the last chunk to start being uploaded
@@ -280,6 +285,8 @@
         byteArrayPool.clear()
     }
 
+    private var failFirstChunkForTesting = true
+
     private fun getReusableByteArray(
         byteArrayPool: ArrayBlockingQueue<ByteArray>,
         inputStream: BufferedInputStream,

```

app/src/main/java/com/infomaniak/swisstransfer/upload/TransferUploader.kt
```diff
@@ -164,52 +182,87 @@ class TransferUploader(
...
-        totalChunks: Int,
-        onUploadBytes: suspend (Long) -> Unit,
     ) {
-        val requestSemaphore = Semaphore(parallelChunks)
+
+        val chunkConfig = metadata.chunkConfig
+        val chunkSize = chunkConfig.fileChunkSize
+        val totalChunks = chunkConfig.totalChunks
+        val parallelChunks = chunkConfig.parallelChunks
+
```
FabianDevel, line 212: These definitions aren't really useful 
You could either make the function an extension of the metaData class (and move it in an extension File along with the other function up there), and you could also put the function as a with(chunkConfig) to already simplify this

app/src/main/java/com/infomaniak/swisstransfer/upload/TransferUploader.kt
```diff
@@ -164,52 +182,87 @@ class TransferUploader(
...
+                        chunkIndex = chunkIndex,
+                        chunkSize = metadata.chunkConfig.fileChunkSize
+                    )
+                    metadata.chunksUploadStatus[chunkIndex] = if (exists) DefinitelyComplete else null
+                }
+            }
+        }
+    }
+
+    private suspend fun uploadFileInChunks(
```
FabianDevel, line 202: This function should be simplified and broken down into smaller ones

FabianDevel (changes requested):
This is better than the actual behaviour, but there is still an inconsistency that the upload retry between 1 and 4 chunks because of the parallelism. 

Most of the time when I tested it, we retry to the closest 200MB

app/src/main/java/com/infomaniak/swisstransfer/upload/UploadForegroundService.kt
```diff
@@ -219,40 +219,37 @@ class UploadForegroundService : ForegroundService(Companion, redeliverIntentIfKi
...
+                                awaitCancellation()
+                            }
                         ) {
-                            uploader.uploadAllOrThrow(updateState = { currentState = it })
+                            withPartialWakeLock(
+                                appName = "SwissTransfer",
+                                tagSuffix = "upload"
+                            ) {
+                                uploader.uploadRemainderOrThrow(updateState = { currentState = it })
+                            }
```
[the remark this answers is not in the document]
LouisCAD: Handled in https://github.com/Infomaniak/android-SwissTransfer/pull/476/files

PR #480, opened 2025-04-17
fix: Make Flow.rateLimit operator fully correct

LouisCAD:
Before, `lastEmitElapsedNanos` wasn't updated if a new value was coming before `send(…)` returned, leading to the sending of the value, and eventual notification being sent, to not be registered, and causing double sending instead of proper rate limiting.

This was a known issue, and commit has the solution.


LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/upload/UploadForegroundService.kt
```diff
@@ -203,7 +206,42 @@ private fun <T> Flow<T>.rateLimit(minInterval: Duration): Flow<T> = channelFlow
         val timeSinceLastEmit = nanosSinceLastEmit.nanoseconds
         val timeToWait = minInterval - timeSinceLastEmit.coerceAtMost(minInterval)
         delay(timeToWait)
-        send(newValue) // We don't handle the case where cancellation happens after it's received.
+        sendAtomic(newValue)
```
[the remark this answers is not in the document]
LouisCAD: Added some doc to make it crystal clear.

app/src/main/java/com/infomaniak/swisstransfer/upload/UploadForegroundService.kt
```diff
@@ -203,7 +206,56 @@ private fun <T> Flow<T>.rateLimit(minInterval: Duration): Flow<T> = channelFlow
...
+        // and throw a `CancellationException` instead of returning (if coroutine's `Job` gets cancelled).
+        sendAtomic(newValue)
         lastEmitElapsedNanos = SystemClock.elapsedRealtimeNanos()
     }
 }.buffer(Channel.RENDEZVOUS)
+
+
+/**
+ * [SendChannel.send] can throw a [CancellationException] after the value was sent,
+ * which might not be desirable if we want to take factor-in whether the value was actually
```
LunarX, line 222: You should choose between "take" and "factor-in"

LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/upload/UploadForegroundService.kt
LouisCAD: Nice catch, removed "take" that was a leftover from rewriting.

LunarX (approved):

PR #492, opened 2025-05-05
chore: Fix duplicate sections

tevincent:
depends on https://github.com/Infomaniak/android-SwissTransfer/pull/494

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -43,7 +43,7 @@ jobs:
 
       # Making sure the app compiles
       - name: Build app
-        run: ./gradlew build
+        run: ./gradlew assembleDebug
 
       - name: Run unit tests
-        run: ./gradlew testDebugUnitTest --stacktrace
+        run: ./gradlew testProdDebugUnitTest --stacktrace
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -41,5 +41,9 @@ jobs:
       - name: Clean gradle cache
         run: ./gradlew clean
 
+      # Making sure the app compiles
+      - name: Build app
+        run: ./gradlew build
+
       - name: Run unit tests
         run: ./gradlew testDebugUnitTest --stacktrace
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -74,7 +74,7 @@ android {
     }
 
     productFlavors {
-        // Don't forget to update the command to build the app in android.yml if necessary
+        // Remember to update the command to build the app in android.yml to build the new flavor when you add a new flavor here
         create("preprod") {
             dimension = "env"
             applicationIdSuffix = ".preprod"
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -74,6 +74,7 @@ android {
     }
 
     productFlavors {
+        // Don't forget to update the command to build the app in android.yml if necessary
         create("preprod") {
             dimension = "env"
             applicationIdSuffix = ".preprod"
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transfers/TransfersGroupingManager.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transfers/TransfersGroupingManager.kt
@@ -22,13 +22,10 @@ import androidx.annotation.StringRes
 import com.infomaniak.multiplatform_swisstransfer.common.interfaces.ui.TransferUi
 import com.infomaniak.swisstransfer.R
 import com.infomaniak.swisstransfer.ui.screen.main.transfers.TransfersGroupingManager.TransferSectionWithContains.Companion.isIn
-import java.time.DayOfWeek
-import java.time.Instant
-import java.time.LocalDate
-import java.time.ZoneId
+import java.time.*
 import java.time.format.TextStyle
 import java.time.temporal.TemporalAdjusters
-import java.util.Locale
+import java.util.*
 
 object TransfersGroupingManager {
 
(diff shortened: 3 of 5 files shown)


.github/workflows/android.yml
```diff
@@ -41,5 +41,9 @@ jobs:
       - name: Clean gradle cache
         run: ./gradlew clean
 
+      # Making sure the app compiles
+      - name: Build app
+        run: ./gradlew buildProdDebug
```
LunarX, line 46: The task seems to have failed because it's not recognized as a valid task

LunarX:
Would it be hard to fix the sentry error instead of changing to `assembleProdDebug`?

.github/workflows/android.yml
```diff
@@ -41,5 +41,9 @@ jobs:
       - name: Clean gradle cache
         run: ./gradlew clean
 
+      # Making sure the app compiles
+      - name: Build app
+        run: ./gradlew assembleProdDebug
+
       - name: Run unit tests
         run: ./gradlew testDebugUnitTest --stacktrace
```
LunarX, line 49: Didn't we need to change this command so the tests would run? Otherwise they weren't?

app/build.gradle.kts
```diff
@@ -77,6 +77,7 @@ android {
     }
 
     productFlavors {
+        // Don't forget to update the command to build the app in android.yml if necessary
```
LunarX, line 80:
```suggestion
        // Don't forget to update the command to build the app in android.yml to build the new flavor when you add a new flavor here
```

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transfers/TransfersGroupingManager.kt
```diff
@@ -95,11 +95,21 @@ object TransfersGroupingManager {
...
+    fun Long.toStartOfMonthInSeconds(): Long {
+        return Instant
+            .ofEpochSecond(this)
+            .atZone(ZoneId.systemDefault())
+            .withDayOfMonth(1)
+            .withSecond(0)
+            .withMinute(0)
+            .withHour(0)
+            .toEpochSecond()
+    }
```
LunarX, line 107: Do you really need this?

app/src/main/res/values-fr/strings.xml
```diff
@@ -101,7 +101,7 @@
     <string name="receivedFilesTitle">Fichiers reçus</string>
     <string name="receivedTitle">Reçu</string>
     <plurals name="recipientHeader">
-        <item quantity="one">Destinataire :</item>
+        <item quantity="one" tools:ignore="ImpliedQuantity">Destinataire :</item>
```
LunarX, line 104: Are all of these changes really needed for your fix?

app/src/test/java/com/infomaniak/swisstransfer/TransferSectionGroupingTest.kt
```diff
@@ -79,14 +81,15 @@ class TransferSectionGroupingTest {
 
     @Test
     fun sections_areOrderCorrectly() {
-        val decemberSection = TransferSection.ByMonth(getMonthNameFromEpoch(december.first()))
+        val yearMonth = december.first().pairOfYearMonth()
+        val decemberTimestamp = TransferSection.ByMonth(yearMonth.first, yearMonth.second)
```
LunarX, line 85: I think the december section should not be called decemberTimestamp

app/src/test/java/com/infomaniak/swisstransfer/TransferSectionGroupingTest.kt
```diff
@@ -79,14 +81,15 @@ class TransferSectionGroupingTest {
...
         assertEquals(groupedTransferUi.keys.elementAt(0), TransferSectionWithContains.Future)
         assertEquals(groupedTransferUi.keys.elementAt(1), TransferSectionWithContains.Today)
         assertEquals(groupedTransferUi.keys.elementAt(2), TransferSectionWithContains.Yesterday)
         assertEquals(groupedTransferUi.keys.elementAt(3), TransferSectionWithContains.ThisWeek)
         assertEquals(groupedTransferUi.keys.elementAt(4), TransferSectionWithContains.LastWeek)
         assertEquals(groupedTransferUi.keys.elementAt(5), TransferSectionWithContains.ThisMonth)
-        assertEquals(groupedTransferUi.keys.elementAt(6), decemberSection)
+        assertEquals((groupedTransferUi.keys.elementAt(6) as TransferSection.ByMonth), decemberTimestamp)
```
LunarX, line 92: I don't think you need this cast as TransferSection.ByMonth

app/src/test/java/com/infomaniak/swisstransfer/TransferSectionGroupingTest.kt
```diff
@@ -106,13 +109,13 @@ class TransferSectionGroupingTest {
...
-                assertEquals(1, constructor?.parameters?.count())
+                assertEquals(2, constructor?.parameters?.count())
 
-                val firstInstanceUid = constructor!!.call("aaa").uid
+                val firstInstanceUid = constructor!!.call(0, 0).uid
                 assert(firstInstanceUid !in previousUids)
                 previousUids.add(firstInstanceUid)
 
-                val secondInstanceUid = constructor.call("bbb").uid
+                val secondInstanceUid = constructor.call(1, 1).uid
```
LunarX, line 118: maybe, just to be sure, we could make it (1, 1) and (2, 2). Just because (0, 0) is actually illegal in the code and it would be a bother if we break a test in the futur by implementing logic that breaks when (0, 0) is provided even though 0 is only provided inside of a test case

LunarX (changes requested):

LunarX (approved):

PR #496, opened 2025-05-12
chore: Add missingDimensionStrategy

TommyDL-Infomaniak:


app/build.gradle.kts
```diff
@@ -76,19 +76,17 @@ android {
     productFlavors {
         // Remember to update the command to build the app in android.yml to build the new flavor when you add a new flavor here
         create("preprod") {
-            dimension = "env"
             applicationIdSuffix = ".preprod"
             buildConfigField("String", "BASE_URL", "\"https://$preprodHost\"")
         }
         create("prod") {
-            dimension = "env"
```
LouisCAD, line 84: Why did you remove these?
According to [the documentation](https://developer.android.com/build/build-variants#product-flavors), it's still needed:
> Error: All flavors must now belong to a named flavor dimension.
> The flavor 'flavor_name' is not assigned to a flavor dimension.

TommyDL-Infomaniak: It is useless. From the documentation you have linked: 
> If you are using only one dimension, this property is optional

FabianDevel (approved):

app/build.gradle.kts
LouisCAD: Ok, so doc is contradicting the code comment in the snippet…
Anyway, if it builds, that's all good, not much risks for that.

TommyDL-Infomaniak:
The changes made in this PR are now included in this one: https://github.com/Infomaniak/android-SwissTransfer/pull/497

PR #509, opened 2025-06-04
feat: Auto add first pasted text that is a valid email

MorganeFar:
Automatically add in a chip the first text that is a valid email found in the clipboard when the user try to paste.

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/pickfiles/components/EmailAddressTextField.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/pickfiles/components/EmailAddressTextField.kt
@@ -17,7 +17,6 @@
  */
 package com.infomaniak.swisstransfer.ui.screen.newtransfer.pickfiles.components
 
-import android.content.ClipboardManager
 import androidx.compose.animation.animateColorAsState
 import androidx.compose.foundation.interaction.MutableInteractionSource
 import androidx.compose.foundation.layout.*
@@ -47,7 +46,6 @@ import com.infomaniak.swisstransfer.ui.theme.Margin
 import com.infomaniak.swisstransfer.ui.theme.SwissTransferTheme
 import com.infomaniak.swisstransfer.ui.utils.GetSetCallbacks
 import com.infomaniak.swisstransfer.ui.utils.PreviewLightAndDark
-import splitties.systemservices.clipboardManager
 
 @OptIn(ExperimentalLayoutApi::class)
 @Composable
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/pickfiles/components/EmailAddressTextField.kt
```diff
@@ -152,14 +159,26 @@ private class EmailAddressTextFieldState(
...
+        val description = clipboardManager.primaryClipDescription ?: return null
+
+        (0 until countItemInClipboard).forEach { index ->
+            if (description.getMimeType(index).startsWith("text/")) {
+                val text = clipboardManager.primaryClip?.getItemAt(index)?.text?.toString()
+                if (text?.isNotBlank() == true) return text
+            }
+        }
+        return null
+    }
```
LunarX, line 173: You should avoid duplicating logic like this and instead find a way to share the logic across all places that need it

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/pickfiles/components/EmailAddressTextField.kt
LunarX, line 1: I see that the code detects email addresses in the clipboard and pastes them automatically for the user, but wasn't 
 the goal of this task different? The goal should be to automatically turn an email address into a chip after the user manually pasted it from its clipboard, right?

LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/pickfiles/components/EmailAddressTextField.kt
```diff
@@ -140,26 +140,36 @@ private class EmailAddressTextFieldState(
     fun updateUiTextValue(newValue: TextFieldValue) {
         var hasNewValidRecipientEmail = false
         val lastAddedChar = newValue.text.lastOrNull()
+        if (textFieldValue.text.length + 1 < newValue.text.length) {
+            hasNewValidRecipientEmail = addRecipientAddress(newValue.text)
+        }
 
         if (lastAddedChar == ' ' || lastAddedChar == ',') {
             hasNewValidRecipientEmail = addRecipientAddress()
         }
```
LunarX, line 149: Both conditions should be mutually exclusive so we don't call `addRecipientAddress` twice in a row if both conditions are met

LunarX (changes requested):

PR #517, opened 2025-07-01
fix: Fix navbar color on theme change

LouisCAD:


app/src/main/java/com/infomaniak/swisstransfer/ui/theme/Theme.kt
```diff
@@ -41,6 +45,12 @@ fun SwissTransferTheme(
     content: @Composable () -> Unit,
 ) {
     val customColors = if (isDarkTheme) CustomDarkColorScheme else CustomLightColorScheme
+    val activity = LocalActivity.current
+    LaunchedEffect(isDarkTheme) {
+        val window = activity?.window ?: return@LaunchedEffect
+        val insetsController = WindowCompat.getInsetsController(window, window.decorView)
+        insetsController.isAppearanceLightNavigationBars = isDarkTheme.not()
+    }
```
LunarX, line 53: There used to be a way to also set the color of the three icons (back, home and recent activities). Can we also set this? Because right now, the background color of the navigation bar is indeed fixed but the color of the icons still adapt to the phone night mode even when the app settings are set to a specific one

PR #518, opened 2025-07-01
feat: Add in-app update required

TommyDL-Infomaniak:
Depends on https://github.com/Infomaniak/android-core/pull/411


app/src/main/java/com/infomaniak/swisstransfer/ui/MainActivity.kt
```diff
@@ -63,12 +73,20 @@ class MainActivity : ComponentActivity(), AppReviewManageable {
...
         super.onCreate(savedInstanceState)
 
         initAppReviewManager()
+        initAppUpdateManager()
+
+        NetworkConfiguration.init(
+            appId = BuildConfig.APPLICATION_ID,
+            appVersionName = BuildConfig.VERSION_NAME,
+            appVersionCode = BuildConfig.VERSION_CODE,
+        )
```
LouisCAD, line 89: Since `MainActivity` isn't the only entry-point of the app, this could be skipped.
Submitted #527 to fix this.

PR #618, opened 2026-04-01
feat: Download a transfer or folder with api v2

sirambd:


app/src/main/res/values/strings.xml
```diff
@@ -33,6 +32,7 @@
     <string name="buttonDownloaded">Downloaded</string>
     <string name="buttonDownloading">Downloading</string>
     <string name="buttonEditTransfer">Edit my transfer</string>
+    <string name="buttonFeedback">Feedback</string>
```
LunarX: Do you need this `buttonFeedback`?

app/src/main/java/com/infomaniak/swisstransfer/ui/utils/NotificationsUtils.kt
```diff
@@ -132,6 +139,11 @@ class NotificationsUtils @Inject constructor(
         notification.post(Ids.LastUpload)
     }
 
+    fun downloadSucceeded(tag: String, title: String) {
+        val notification = buildDownloadNotification(title)
+        notificationManagerCompat.notifyCompat(tag, 1, notification)
```
LunarX: This method looks deprecated

app/src/main/java/com/infomaniak/swisstransfer/ui/utils/NotificationsUtils.kt
```diff
@@ -155,6 +167,30 @@ class NotificationsUtils @Inject constructor(
...
+            .setOngoing(true)
+            .setForegroundServiceBehavior(NotificationCompat.FOREGROUND_SERVICE_IMMEDIATE)
+            .setSmallIcon(android.R.drawable.stat_sys_download)
+            .setColor(notificationIconColor)
+            .build()
+    }
+
+    private fun buildDownloadNotification(title: String): Notification {
+        return NotificationCompat.Builder(appContext, ChannelIds.downloadChannelId)
+            .setTicker(title)
```
LunarX: Shouldn't the ticker be "download completed" rather than the title of the file? Or if you want to go further maybe a mix of the two like "Download completed: $title" but this requires a new string

app/src/main/java/com/infomaniak/swisstransfer/ui/utils/NotificationsUtils.kt
```diff
@@ -155,6 +167,30 @@ class NotificationsUtils @Inject constructor(
...
+            .setContentTitle(title)
+            .setContentText(appContext.getString(R.string.notificationDownloadSuccessNotificationTitle))
+            .setSmallIcon(defaultSmallIcon)
+            .setColor(notificationIconColor)
+            .setAutoCancel(true)
+            .build()
+    }
```
LunarX: Do we not put an intent when the notification is clicked?

app/src/main/java/com/infomaniak/swisstransfer/services/DownloadWorker.kt
```diff
@@ -0,0 +1,268 @@
...
+            val inputData = workDataOf(TRANSFER_ID_KEY to transferId, FOLDER_ID_KEY to folderId)
+            val workRequest = OneTimeWorkRequest.Builder(DownloadWorker::class.java)
+                .setInputData(inputData)
+                .setConstraints(constraints)
+                .build()
+            workManager.enqueueUniqueWork(uniqueWorkName, ExistingWorkPolicy.REPLACE, workRequest)
+            return UniqueDownloadId(999)
+        }
+
+        fun downloadStatusFlow(transferId: String, folderId: String?): Flow<DownloadStatus> {
```
LunarX, line 228: Is this really a folderId or rather a fileId?

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/TransferDownload.kt
```diff
@@ -81,20 +86,32 @@ suspend fun handleTransferDownload(
     transferManager = transferManager,
     ui = ui,
     apiUrlCreator = apiUrlCreator,
+    downloadWorkerScheduler = downloadWorkerScheduler,
     userAgent = userAgent,
     transfer = transfer,
     targetFile = targetFile,
     direction = direction,
 ).collectLatest { id ->
+    val isNeedDownloadWorker = transfer.isV2() && targetFile?.isFolder != false
```
LunarX:
```suggestion
    val needsDownloadWorker = transfer.isV2() && targetFile?.isFolder != false
```

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/TransferDownload.kt
```diff
@@ -146,34 +171,44 @@ private fun currentOrNewDownloadManagerId(
...
 }.first()
 
-private suspend fun isFileDeleted(id: UniqueDownloadId): Boolean {
+private suspend fun isFileDeleted(id: UniqueDownloadId, transfer: TransferUi, targetFile: FileUi?): Boolean {
     repeat(2) { attemptIndex ->
         runCatching {
-            return downloadManager.doesFileExist(id).not()
+            return when {
+                transfer.isV1() -> downloadManager.doesFileExist(id).not()
+                targetFile != null -> AppDownloadManager.doesFileExists(transfer, targetFile).not()
```
LunarX:
```suggestion
                targetFile != null -> AppDownloadManager.doesFileExist(transfer, targetFile).not()
```

app/src/main/java/com/infomaniak/swisstransfer/services/AppDownloadManager.kt
```diff
@@ -0,0 +1,277 @@
...
+import splitties.init.appCtx
+import java.io.File
+import java.io.FileOutputStream
+import java.io.OutputStream
+import javax.inject.Inject
+import kotlin.concurrent.atomics.AtomicLong
+import kotlin.concurrent.atomics.ExperimentalAtomicApi
+
+class AppDownloadManager @Inject constructor(
+    @param:ApplicationContext private val appContext: Context,
```
LunarX, line 63: Wouldn't we want to apply ApplicationContext to the field as well as the parameter?

app/src/main/java/com/infomaniak/swisstransfer/services/AppDownloadManager.kt
```diff
@@ -0,0 +1,277 @@
...
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            downloadToPublicDownloads(
+                url = downloadUrl,
+                fileUi = fileUi,
+                path = path,
+                onProgress = onProgress,
+            )
+        } else {
+            downloadToPublicDownloadsUnderApi29(
```
LunarX:
```suggestion
            downloadToPublicDownloadsBelowApi29(
```

app/src/main/java/com/infomaniak/swisstransfer/services/AppDownloadManager.kt
```diff
@@ -0,0 +1,277 @@
...
+        private const val DEFAULT_BUFFER_SIZE: Long = 8 * 1024
+        const val ROOT_FOLDER_NAME = "SwissTransfer"
+
+        private const val PARALLEL_DOWNLOADS_COUNT = 6
+
+        fun TransferUi.computeFolderDownloadPathWith(fileUi: FileUi): String {
+            val filePath = fileUi.path?.takeIf { it.contains("/") }?.let {
+                if (it.startsWith("/")) it.substringAfter("/") else it
+            }
+            return "$ROOT_FOLDER_NAME/${this.displayTitle}/${filePath?.substringBeforeLast("/") ?: ""}"
```
LunarX: Can't we just do something like `fileUi.path?.removeSurrounding("/")`?

app/src/main/java/com/infomaniak/swisstransfer/services/AppDownloadManager.kt
```diff
@@ -0,0 +1,277 @@
...
+        }
+
+        suspend fun uriFor(transferUi: TransferUi, fileUi: FileUi): Uri? = withContext(Dispatchers.IO) {
+            return@withContext when {
+                Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q -> uriForAboveApi29(transferUi, fileUi)
```
LunarX: We always write it `SDK_INT >= 29` elsewhere. Also I just realized you have the same issue above in your code as well

app/src/main/java/com/infomaniak/swisstransfer/services/AppDownloadManager.kt
```diff
@@ -0,0 +1,277 @@
...
+
+        @RequiresApi(Build.VERSION_CODES.Q)
+        private fun uriForAboveApi29(transferUi: TransferUi, fileUi: FileUi): Uri? {
+            val contentUri = MediaStore.Downloads.EXTERNAL_CONTENT_URI
+            val relativePath = "${Environment.DIRECTORY_DOWNLOADS}/${transferUi.computeFolderDownloadPathWith(fileUi)}"
+            val selection = "${MediaStore.DownloadColumns.RELATIVE_PATH}=? AND ${MediaStore.DownloadColumns.DISPLAY_NAME}=?" +
+                    " AND ${MediaStore.DownloadColumns.IS_PENDING}=0"
+            val selectionArgs = arrayOf(relativePath, fileUi.fileName)
+
+            return appCtx.contentResolver.query(contentUri, null, selection, selectionArgs, null)?.use { cursor ->
```
LunarX: You once told me it's was best to not use a project for compatibility reasons, that's it right?

app/src/main/java/com/infomaniak/swisstransfer/services/DownloadWorker.kt
```diff
@@ -0,0 +1,285 @@
...
+                    appDownloadManager.downloadTransferToPublicDownload(transferUi) { downloadedBytes, totalBytes ->
+                        updateForegroundNotificationIfNeeded(title, downloadedBytes, totalBytes)
+                    }
+
+                    notificationsUtils.downloadSucceeded(tag = uniqueWorkName(transferId, null), title)
+                }
+            }
+        }.getOrElse { exception ->
+            return when (exception) {
+                is CancellationException, is NetworkException -> Result.retry()
```
LunarX, line 100: Why do we want to retry when there's a CancellationException?

app/src/main/java/com/infomaniak/swisstransfer/services/DownloadWorker.kt
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    private fun hasAvailableSpace(requiredBytes: Long): Boolean {
+        val path = when {
+            Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q -> applicationContext.getExternalFilesDir(null)?.path ?: return true
+            else -> Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS).path
+        }
+        val availableBytes = StatFs(path).availableBytes
+        // Add a safety margin of 10% extra space
+        val requiredWithMargin = (requiredBytes * 110) / 100
```
LunarX:
```suggestion
        val requiredWithMargin = requiredBytes * 1.1
```

app/src/main/java/com/infomaniak/swisstransfer/services/DownloadWorker.kt
```diff
@@ -0,0 +1,285 @@
...
+                /* notificationId = */ foregroundId,
+                /* notification = */
+                notificationsUtils.buildDownloadProgressNotification(title, percent.toInt(), indeterminate = totalBytes == 0L),
+            )
+        }
+    }
+
+    private fun hasAvailableSpace(requiredBytes: Long): Boolean {
+        val path = when {
+            Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q -> applicationContext.getExternalFilesDir(null)?.path ?: return true
```
LunarX: Same writing standard
```suggestion
           SDK_INT >= 29 -> applicationContext.getExternalFilesDir(null)?.path ?: return true
```

app/src/main/java/com/infomaniak/swisstransfer/services/DownloadWorker.kt
```diff
@@ -0,0 +1,285 @@
...
+            return workManager.getWorkInfosFlow(workQuery).transform {
+                val workInfo = it.firstOrNull()
+                if (workInfo == null) {
+                    emit(DownloadStatus.Complete)
+                } else if (workInfo.isPending() && isNetworkAvailableFlow.first().not()) {
+                    emit(DownloadStatus.Paused(DownloadStatus.Paused.Reason.WaitingForNetwork))
+                } else {
+                    emit(workInfo.toDownloadStatus())
+                }
+            }
```
LunarX: Why not just use a map instead of a transform?

app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItem.kt
LunarX: As sonar pointed out, createdDate is unused

LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/transferdetails/TransferDownload.kt
```diff
@@ -146,34 +171,46 @@ private fun currentOrNewDownloadManagerId(
...
             transfer = transfer,
             targetFile = targetFile,
             direction = direction,
         ) ?: return@repeatWhileActive
     }
-}.distinctUntilChanged()
+}.apply {
+    // Since api v2 the download id of a folder is always the same (see: DownloadWork.DOWNLOAD_WORKER_DOWNLOAD_ID)
+    if (transfer.isV1() || targetFile?.isFolder == false) distinctUntilChanged()
+}
```
LunarX:
```suggestion
}.let {
    // Since api v2 the download id of a folder is always the same (see: DownloadWork.DOWNLOAD_WORKER_DOWNLOAD_ID)
    if (transfer.isV1() || targetFile?.isFolder == false) it.distinctUntilChanged() else it
}
```

LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/services/DownloadWorker.kt
sirambd: It's a folderId

app/src/main/java/com/infomaniak/swisstransfer/services/AppDownloadManager.kt
sirambd: @param: applies the annotation to the constructor parameter, which is what is intended here, for @field, it applies only to the backing field, so HILT might be lost because here the backing field will be in the generated code and not in the constructor

app/src/main/java/com/infomaniak/swisstransfer/services/AppDownloadManager.kt
sirambd: Finally I used **Before**, and now it's `downloadToPublicDownloadsBeforeApi29`

app/src/main/java/com/infomaniak/swisstransfer/services/AppDownloadManager.kt
sirambd: No we can't because the documentation clearly states that it removes the `delimiter` only if the string begins and ends with the delimiter.
Here we only want to remove the delimiter from the beginning when necessary.

app/src/main/java/com/infomaniak/swisstransfer/services/AppDownloadManager.kt
sirambd: Project ??? Do you mean Projection ?

sirambd: If we're talking about projections, then I'd say it depends on which fields we might need, since they might not actually exist.
But here we only need one column, so we can include it to optimize the Mediastore query.
In fact, this has already been done in commit 83d331a3061e99d97c03c4900041a0c44d311491

app/src/main/java/com/infomaniak/swisstransfer/services/DownloadWorker.kt
sirambd: Given that, in a Worker, CancellationExceptions mainly originate from the system, and generally serve to tell us to stop and restart later, and since we don’t perform any jobs that we cancel ourselves, we don’t encounter CancellationExceptions that apply to us and need to be handled, because all our errors are exceptions that provide clear information.

app/src/main/java/com/infomaniak/swisstransfer/ui/components/transfer/TransferItem.kt
sirambd: And he's right, because it's unnecessary now that we have the `displayTitle` attribute

app/src/main/java/com/infomaniak/swisstransfer/services/DownloadWorker.kt
sirambd: Looking at the code, I don't see any particular reason; clearly, a `map` would suffice here.
I'd even say a `mapLatest`, since we have a suspend function.
Check out commit 6b6f9932986cd590f45e51b59d719cdbd70a37f0

app/src/main/java/com/infomaniak/swisstransfer/services/AppDownloadManager.kt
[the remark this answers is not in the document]
sirambd: We already know what the API returns, so there's no need to go that far

app/src/main/java/com/infomaniak/swisstransfer/ui/utils/TransferUiExt.kt
```diff
@@ -85,6 +86,13 @@ val TransferUi.isExpired: Boolean
         return expiry.before(now) || downloadLeft <= 0
     }
 
+val TransferUi.displayTitle: String
+    get() {
+        val createdDate = createdDateTimestamp.toDateFromSeconds().format(FORMAT_DATE_TITLE)
+        val title = title ?: createdDate
+        return title
```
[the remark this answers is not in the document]
sirambd: @copilot Check it out—I've improved this code even further in this commit: 075b16f19b30ea149670e4655c88dc810bee7597

app/src/main/java/com/infomaniak/swisstransfer/ui/utils/NotificationsUtils.kt
sirambd: @copilot There's no point in displaying the download view because they'll never end up there, instead, `packageManager.getLaunchIntentForPackage` would be better here.

LunarX (approved):

PR #631, opened 2026-04-30
fix: Avoid reprocessing API v2 deeplink multiple times

sirambd:
**Description**
Currently, when you have an APIv2 deep link with a folder, as long as you navigate within the folder and return to the root transfer, you can duplicate the data because, as long as the intent is active and you return to the root of a transfer, the transfer is re-added to the database each time you return.

**Fix**
Handle API v2 deeplink flow correctly to prevent duplicate processing
while the intent is still active.


LunarX (changes requested):
When I try to deeplink to a transfer it fails to open it so I can't test if the bug is fixed

LunarX (approved):

PR #639, opened 2026-05-06
feat: Protect logout button with an AlertDialog

sirambd:


app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/MyAccountScreen.kt
```diff
@@ -267,13 +273,14 @@ private fun SettingsItems(
...
-            SettingItem(
-                titleRes = R.string.settingsLogOut,
-                isSelected = { selectedSetting == Logout },
-                icon = AppIcons.DoorRectangleArrowRight,
-                onClick = { onActionAndMatomo(Logout) },
-            )
+        AnimatedVisibility(currentUser != null) {
+            currentUser?.let {
+                LogoutItem(
+                    currentUser = it,
```
LunarX: Only pass email directly to the item I'd say
```suggestion
            currentUser?.let {
                LogoutItem(
                    email = it.email,
```

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/settings/MyAccountScreen.kt
```diff
@@ -267,13 +273,14 @@ private fun SettingsItems(
...
+                    isSelected = { selectedSetting == Logout },
+                    onConfirmClick = { onActionAndMatomo(Logout) },
+                )
+            }
         }
```
LunarX, line 282: Replace AnimatedVisibility with AnimatedContent to access the non-null `currentUser` during the transition or provide a nullable email to `LogoutItem` to avoid the `currentUser?.let {}` which will prevent the animation from playing out correctly by removing the LougoutItem from the composition too early

app/src/main/res/values/strings.xml
```diff
@@ -38,6 +38,8 @@
     <string name="buttonStart">Start</string>
     <string name="cantImportFileX">Can’t import the file %s</string>
     <string name="checkingFiles">Checking files…</string>
+    <string name="confirmLogoutDescription">Are you sure you want to log out from %s?</string>
+    <string name="confirmLogoutTitle">Log out</string>
```
LunarX: Is the key `settingsLogOut` not good enough?

LunarX (changes requested):

LunarX (changes requested):
The key is still present on the remote

LunarX (approved):

PR #645, opened 2026-05-18
feat: Add transition

NicolasBourdin88:
This PR centralizes the application's transition logic into a single place: `SwissTransferTransition.kt`.
The goal is to make transitions easier to maintain and update, since changing this file is now enough to affect transitions across the app.

It also introduces shared TopAppBar transition handling to prevent the blinking that previously occurred when navigating between screens.
Before this change, each screen defined its own TopAppBar inside its Scaffold, so during screen transitions the app switched between different app bars, causing visible flickering.

In addition, this PR introduces the concept of a TopBarKey.
This is especially important for tablet mode, where up to three different TopAppBars can be displayed at the same time and need to be identified and transitioned correctly.

Depends on https://github.com/Infomaniak/android-SwissTransfer/pull/647

### DIFF
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTransition.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTransition.kt
@@ -22,14 +22,13 @@ import androidx.compose.animation.SharedTransitionScope
 import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.remember
 import androidx.compose.runtime.staticCompositionLocalOf
 import androidx.compose.ui.Modifier
 
 val LocalSharedTransitionScope = staticCompositionLocalOf<SharedTransitionScope?> { null }
 val LocalAnimatedPaneVisibilityScope = staticCompositionLocalOf<AnimatedVisibilityScope?> { null }
 val LocalNavHostAnimatedVisibilityScope = staticCompositionLocalOf<AnimatedVisibilityScope?> { null }
-val LocalCurrentTopBarKey = staticCompositionLocalOf { TopBarKey.FIRST }
+val LocalCurrentTopBarKey = staticCompositionLocalOf { TopBarKey.SINGLE_PANE }
 
 object SwissTransferTransition {
     val enterTransition = fadeIn()
@@ -43,10 +42,8 @@ fun Modifier.sharedTransitionAppBar(): Modifier {
     val navHostScope = LocalNavHostAnimatedVisibilityScope.current
     val currentTopBarKey = LocalCurrentTopBarKey.current
 
-    val activeScope = remember(navHostScope, localScope) {
-        val isNavHostAnimating = navHostScope?.transition?.let { it.currentState != it.targetState } == true
-        if (isNavHostAnimating) navHostScope else localScope
-    }
+    val isNavHostAnimating = navHostScope?.transition?.let { it.currentState != it.targetState } == true
+    val activeScope = if (isNavHostAnimating) navHostScope else localScope
 
     if (activeScope == null || sharedScope == null) return this
 
@@ -59,5 +56,5 @@ fun Modifier.sharedTransitionAppBar(): Modifier {
 }
 
 enum class TopBarKey {
-    FIRST, SECOND, THIRD
+    SINGLE_PANE, LIST_PANE, DETAIL_PANE
 }
--- a/app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTransition.kt
+++ b/app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTransition.kt
@@ -1,0 +1,31 @@
+/*
+ * Infomaniak SwissTransfer - Android
+ * Copyright (C) 2026 Infomaniak Network SA
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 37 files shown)


app/src/main/java/com/infomaniak/swisstransfer/ui/utils/SwissTransferTransition.kt
LunarX, line 1: Maybe utils would be a better package than component to store this file

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTransition.kt
```diff
@@ -0,0 +1,60 @@
...
+object SwissTransferTransition {
+    val enterTransition = fadeIn()
+    val exitTransition = fadeOut()
+}
+
+@Composable
+fun Modifier.sharedTransitionAppBar(): Modifier {
+    val sharedScope = LocalSharedTransitionScope.current
+    val localScope = LocalAnimatedPaneVisibilityScope.current
+    val navHostScope = LocalNavHostAnimatedVisibilityScope.current
```
LunarX:
```suggestion
    val localScope = LocalAnimatedPaneVisibilityScope.current ?: return this
    val navHostScope = LocalNavHostAnimatedVisibilityScope.current ?: return this
```

app/src/main/java/com/infomaniak/swisstransfer/ui/components/SwissTransferTransition.kt
```diff
@@ -0,0 +1,60 @@
...
+    return with(sharedScope) {
+        this@sharedTransitionAppBar then Modifier.sharedElement(
+            sharedContentState = rememberSharedContentState(key = currentTopBarKey),
+            animatedVisibilityScope = activeScope,
+        )
+    }
+}
+
+enum class TopBarKey {
+    SINGLE_PANE, LIST_PANE, DETAIL_PANE
```
LunarX: New enums are written in PascalCase

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/MainNavHost.kt
```diff
@@ -48,25 +51,41 @@ fun MainNavHost(
...
+                        TransfersScreenWrapper(
+                            direction = TransferDirection.RECEIVED,
+                            transferUuid = args.transferUuid,
+                            isApiV2Deeplink = args.isApiV2,
+                            hideBottomBar = hideBottomBar,
+                        )
+                    }
+                }
+                composable<MyAccountDestination> {
+                    CompositionLocalProvider(LocalNavHostAnimatedVisibilityScope provides this@composable) {
```
LunarX: Maybe we can factorize the CompositionLocalProvider for LocalNavHostAnimatedVisibilityScope in a custom composable<T> of ours

app/src/main/java/com/infomaniak/swisstransfer/ui/components/TwoPaneScaffold.kt
```diff
@@ -84,12 +90,67 @@ fun <T> TwoPaneScaffold(
...
+        transitionSpec = {
+            SwissTransferTransition.enterTransition togetherWith SwissTransferTransition.exitTransition
+        },
+    ) { targetDestination ->
+        val activeScope = LocalAnimatedPaneVisibilityScope.current?.let { paneScope ->
+            val contentScope = this@AnimatedContent
+
+            remember(paneScope, contentScope) {
+                val isPaneAnimating = paneScope.isAnimating()
+                if (isPaneAnimating) paneScope else contentScope
```
LunarX:
```suggestion
                if (paneScope.isAnimating()) paneScope else contentScope
```

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/received/ReceivedScreen.kt
```diff
@@ -118,24 +121,29 @@ private fun ReceivedContent(
...
-            modifier = Modifier.fillMaxHeight(),
-            direction = TransferDirection.RECEIVED,
-            navigateToDetails = navigateToDetails,
-            getSelectedTransferUuid = getSelectedTransferUuid,
-            getTransfers = { transfers },
-            onDeleteTransfer = onDeleteTransfer,
-        )
+    AnimatedContent(
+        targetState = transfers.isEmpty(),
+        transitionSpec = { SwissTransferTransition.enterTransition togetherWith SwissTransferTransition.exitTransition },
```
LunarX: Try extracting this variable to SwissTransferTransition like the other two values

LunarX (changes requested):

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/received/ReceivedScreen.kt
```diff
@@ -103,7 +103,11 @@ private fun ReceivedScreen(
             if (windowAdaptiveInfo.isWindowSmall()) ReceivedEmptyFab(isMessageVisible = { isFirstTransfer() })
         },
     ) {
-        if (uiState() is TransferUiState.Success) {
+        AnimatedVisibility(
+            visible = uiState() is TransferUiState.Success,
+            enter = SwissTransferTransition.enterTransition,
+            exit = SwissTransferTransition.exitTransition,
```
LunarX: Maybe this doesn't need to rely on our factorized screen transition spec

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/main/received/ReceivedScreen.kt
```diff
@@ -123,7 +127,7 @@ private fun ReceivedContent(
 ) {
     AnimatedContent(
         targetState = transfers.isEmpty(),
-        transitionSpec = { SwissTransferTransition.enterTransition togetherWith SwissTransferTransition.exitTransition },
+        transitionSpec = { SwissTransferTransition.transitionSpec },
```
LunarX: Check all issues there could be between these two methods

LunarX (changes requested):

LunarX (approved):

PR #676, opened 2026-07-07
refactor: Enhance Sentry error logging for UploadState

aymericmariaux:


tevincent (changes requested):
It seems that you just print a static value to sentry.

sirambd (approved):
The rules are fine, but you need to test them locally to make sure they work properly.

tevincent (approved):

PR #679, opened 2026-07-29
feat: Add organization interface

aymericmariaux:
depends on : https://github.com/Infomaniak/android-core/pull/824
depends on : https://github.com/Infomaniak/multiplatform-SwissTransfer/pull/312


app/src/main/java/com/infomaniak/swisstransfer/ui/components/OrganizationSwitcherBottomSheet.kt
```diff
@@ -0,0 +1,191 @@
...
+) {
+    val sheetState = rememberModalBottomSheetState(skipPartiallyExpanded = true)
+    val scope = rememberCoroutineScope()
+
+    SwissTransferBottomSheet(
+        onDismissRequest = closeBottomSheet,
+        modifier = modifier,
+        sheetState = sheetState,
+        title = stringResource(R.string.settingsOptionOrganization),
+        content = {
```
LouisCAD: You can use the trailing lambda parameter syntax here, that will save an indentation level without making the code less clear, which is good since it's already quite deep.

app/src/main/java/com/infomaniak/swisstransfer/ui/components/OrganizationSwitcherBottomSheet.kt
```diff
@@ -0,0 +1,191 @@
...
+        leading = {
+            Avatar(
+                avatarType = organization.avatarType(),
+                modifier = Modifier
+                    .size(CoreDimens.bigAvatarSize)
+                    .border(Dimens.BorderWidth, SwissTransferTheme.materialColors.outline, CustomShapes.EXTRA_SMALL),
+                shape = CustomShapes.EXTRA_SMALL,
+            )
+        },
+        content = {
```
LouisCAD: Same as above. Use the trailing lambda syntax.

app/src/main/java/com/infomaniak/swisstransfer/ui/components/OrganizationSwitcherViewModel.kt
```diff
@@ -0,0 +1,48 @@
...
+    @IoDispatcher private val ioDispatcher: CoroutineDispatcher,
+) : ViewModel() {
+
+    val selectedOrganization: StateFlow<OrganizationAccount?> = accountManager.selectedOrganizationAccount()
+        .stateIn(viewModelScope, SharingStarted.Lazily, initialValue = null)
+
+    val organizations: StateFlow<List<OrganizationAccount>> = accountManager.organizationAccountsForCurrentUser()
+        .stateIn(viewModelScope, SharingStarted.Lazily, initialValue = emptyList())
+
+    fun switchToOrganization(organizationAccountId: Long) = viewModelScope.launch(ioDispatcher) {
```
LouisCAD: Why did you use `ioDispatcher` here? This should not be needed for proper suspend functions.

app/src/main/java/com/infomaniak/swisstransfer/ui/screen/newtransfer/pickfiles/PickFilesScreen.kt
```diff
@@ -523,16 +548,16 @@ private fun PickFilesTitle(@StringRes titleRes: Int, modifier: Modifier = Modifi
 private fun SendButton(
     canSendStatus: () -> CanSendStatus,
     expectsClick: () -> Boolean,
-    onClick: () -> Unit,
     modifier: Modifier = Modifier,
+    onClick: () -> Unit
```
LouisCAD: We don't want `onClick` to be the last parameter, because we don't want it to be used with the unnamed trailing lambda syntax that could be ambiguous with the typical case of having a `content` composable function parameter (like with `Row`, `Column`, and other slot based composables).

LouisCAD (changes requested):
Made a preliminary review. I will do another, more thorough review early next week.

LouisCAD (approved):
Nice job!

